PR #59, opened 2015-01-29
for windows.

daiksy:


util/windows/windows_api.go
```diff
@@ -34,24 +36,34 @@ type MEMORYSTATUSEX struct {
...
 }
 
+// PDH_FMT_COUNTERVALUE_ITEM_DOUBLE XXX
 type PDH_FMT_COUNTERVALUE_ITEM_DOUBLE struct {
 	Name     *uint16
 	FmtValue PDH_FMT_COUNTERVALUE_DOUBLE
 }
 
 const (
+	// ERROR_SUCCESS XXX
```
motemen: You can just add comments to enclosing `const` block to suppress lint, not each one.

util/windows/windows_api.go
```diff
@@ -60,22 +65,39 @@ var (
...
+	GetLastError				= modkernel32.NewProc("GetLastError")
+	// PdhOpenQuery XXX
 	PdhOpenQuery                = modpdh.NewProc("PdhOpenQuery")
+	// PdhAddCounter XXX
 	PdhAddCounter               = modpdh.NewProc("PdhAddCounterW")
+	// PdhCollectQueryData XXX
 	PdhCollectQueryData         = modpdh.NewProc("PdhCollectQueryData")
+	// PdhGetFormattedCounterValue XXX
 	PdhGetFormattedCounterValue = modpdh.NewProc("PdhGetFormattedCounterValue")
+	// PdhCloseQuery XXX
```
motemen: You can omit these one-by-one docs too

util/windows/filesystem.go
```diff
@@ -0,0 +1,90 @@
...
+		uintptr(unsafe.Pointer(&drivebuf[0])))
+	if r == 0 {
+		e := GetLastError.Call
+		windowsLogger.Errorf("error code is  [%q]", e)
+		return nil, err
+	}
+
+	drives := []string{}
+	for _, v := range drivebuf {
+		if v >= 65 && v <= 90 {
```
motemen: Can't we use `'A'` and `'Z'` here?

metrics/windows/memory.go
```diff
@@ -21,18 +20,25 @@ func NewMemoryGenerator() (*MemoryGenerator, error) {
...
-	r, _, err := GlobalMemoryStatusEx.Call(uintptr(unsafe.Pointer(&memoryStatusEx)))
+	r, _, err := windows.GlobalMemoryStatusEx.Call(uintptr(unsafe.Pointer(&memoryStatusEx)))
 	if r == 0 {
 		return nil, err
 	}
 
-	ret["memory.total"] = float64(memoryStatusEx.TotalPhys) / 1024
-	ret["memory.free"] = float64(memoryStatusEx.AvailPhys) / 1024
+	free := float64(memoryStatusEx.AvailPhys) / 1024 * 1000
+	total := float64(memoryStatusEx.TotalPhys) / 1024 * 1000
```
motemen: What does `* 1000` do?

metrics/windows/cpuusage.go
```diff
@@ -3,32 +3,32 @@
...
+	return &CPUUsageGenerator{}, nil
 }
 
 // Generate XXX
 func (g *CPUUsageGenerator) Generate() (metrics.Values, error) {
-	if g == nil {
-		return nil, errors.New("CPUUsageGenerator is not initialized")
-	}
-	time.Sleep(g.Interval * time.Second)
+	cpuusage := make(map[string]float64, 1)
```
motemen: Shouldn't we make a map of capacity=2 here? (or omit it)

metrics/windows/disk.go
```diff
@@ -73,11 +74,8 @@ func NewDiskGenerator(interval time.Duration) (*DiskGenerator, error) {
 		}
 	}
 
-	r, _, err = PdhCollectQueryData.Call(uintptr(g.query))
-	if r != 0 && err != nil {
-		diskLogger.Criticalf(err.Error())
-		return nil, err
-	}
+	windows.PdhCollectQueryData.Call(uintptr(g.query))
```
motemen: Is it okay to discard errors on PdhCollectQueryData?

metrics/windows/filesystem.go
```diff
@@ -0,0 +1,55 @@
...
+
+	ret := make(map[string]float64)
+	for name, values := range filesystems {
+		if matches := regexp.MustCompile(`^(.*):`).FindStringSubmatch(name); matches != nil {
+			device := regexp.MustCompile(`[^A-Za-z0-9_-]`).ReplaceAllString(matches[1], "_")
+
+			for key, value := range values {
+				floatValue, valueTypeOk := value.(float64)
+				if valueTypeOk {
+					if(key == "kb_size") {
```
motemen: gofmt required?

spec/windows/block_device.go
```diff
@@ -7,26 +7,31 @@ import (
...
 	results := make(map[string]map[string]interface{})
 
 	drivebuf := make([]byte, 256)
-	_, r, err := GetLogicalDriveStrings.Call(
+	r, _, err := windows.GetLogicalDriveStrings.Call(
 		uintptr(len(drivebuf)),
 		uintptr(unsafe.Pointer(&drivebuf[0])))
-	if r != 0 {
+	if r == 0 {
+		e := windows.GetLastError.Call
```
motemen: Don't we need `()` to get last error?

I think `err` variable here already has the result `GetLastError` as far as I've read from the doc?: https://github.com/golang/go/blob/e67a39e/src/syscall/dll_windows.go#L121-124

util/windows/filesystem.go
```diff
@@ -0,0 +1,90 @@
...
+	"strings"
+	"syscall"
+	"unsafe"
+	"github.com/mackerelio/mackerel-agent/logging"
+)
+
+var windowsLogger = logging.GetLogger("windows")
+
+// CollectDfValues XXX
+func CollectDfValues() (map[string]map[string]interface{}, error) {
```
motemen: I don't think we should name this function as `Df` as this function does not call `df`; Also we don't need to pass data by `map[string]interface{}`, currently we have type assertion inside `metrics/windows/FilesystemGenerator.Generate()`, but we might pass them by structs.

spec/windows/block_device.go
```diff
@@ -7,26 +7,31 @@ import (
...
 func (g *BlockDeviceGenerator) Generate() (interface{}, error) {
 	results := make(map[string]map[string]interface{})
 
 	drivebuf := make([]byte, 256)
-	_, r, err := GetLogicalDriveStrings.Call(
+	r, _, err := windows.GetLogicalDriveStrings.Call(
 		uintptr(len(drivebuf)),
 		uintptr(unsafe.Pointer(&drivebuf[0])))
-	if r != 0 {
+	if r == 0 {
```
motemen: [IMO] I'd like to rename this variable to other than `r` because this return value is set to 0 on errors, unlike others.

util/windows/filesystem.go
```diff
@@ -0,0 +1,90 @@
...
+			"kb_available": float64(freeBytesAvailable / 1024),
+			"mount":        drive,
+			"label":        syscall.UTF16ToString(drivebuf),
+			"volume_name":  syscall.UTF16ToString(volumebuf),
+			"fs_type":      strings.ToLower(syscall.UTF16ToString(fsnamebuf)),
+		}
+	}
+
+	return filesystems, nil
+}
```
motemen: Oh, the code here was a moved one, I think we may refactor it later.

util/util_windows.go
```diff
@@ -20,3 +24,35 @@ func RunCommand(command string) (string, string, error) {
...
+	if err != nil {
+		return 0, err
+	}
+
+	ret, err := strconv.ParseFloat(value, 64)
+	if err != nil {
+		return 0, err
+	}
+	return ret, nil
+}
```
motemen: [IMO] These two functions may moved under util/windows as they are windows specific.

util/windows/filesystem.go
daiksy: `'A'` is 65.
`'Z'` is 90.

I think we can use `'A'` and `'Z'`.

spec/windows/block_device.go
daiksy: Return code's value is windows specification what `0` is error, or  `0` is not error. 

`r` means `return code`.

I think that it may be a this name.

metrics/windows/memory.go
daiksy: oh, It's mistake! Correctly is `1024`.

spec/windows/block_device.go
motemen: （日本語で失礼します）r って無性格な語であまり意味が無いので、ほかの r と違って 0 が成功でないことがわかってる場合には別の語にしといたほうがいいのじゃないか、という動機でした。この場合は長さを指してると思うので n とか。

daiksy: I understand！ :ok_woman:

metrics/windows/disk.go
```diff
@@ -45,13 +46,13 @@ func NewDiskGenerator(interval time.Duration) (*DiskGenerator, error) {
 	for _, v := range drivebuf {
```
motemen: According to https://msdn.microsoft.com/library/cc429332.aspx, `drivebuf` is a concatenation of null-terminated strings, so we should examine it as strings, not bytes.

Actually this code is not introduced by this PR, so we may fix it later.

metrics/windows/memory.go
```diff
@@ -21,18 +20,25 @@ func NewMemoryGenerator() (*MemoryGenerator, error) {
...
-	r, _, err := GlobalMemoryStatusEx.Call(uintptr(unsafe.Pointer(&memoryStatusEx)))
+	r, _, err := windows.GlobalMemoryStatusEx.Call(uintptr(unsafe.Pointer(&memoryStatusEx)))
 	if r == 0 {
 		return nil, err
 	}
 
-	ret["memory.total"] = float64(memoryStatusEx.TotalPhys) / 1024
-	ret["memory.free"] = float64(memoryStatusEx.AvailPhys) / 1024
+	free := float64(memoryStatusEx.AvailPhys) / 1024 * 1024
+	total := float64(memoryStatusEx.TotalPhys) / 1024 * 1024
```
motemen: I think ` / 1024 * 1024` does nothing?

motemen:
LGTM! :rocket: 

[](http://www.pusheen.com/post/28883730187)

PR #63, opened 2015-02-06
Windows metric brushup

daiksy:


util/util_windows.go
```diff
@@ -26,3 +29,47 @@ func RunCommand(command string) (string, string, error) {
...
+type FilesystemInfo struct {
+	PercentUsed string
+	KbUsed      float64
+	KbSize      float64
+	KbAvailable float64
+	Mount        string
+	Label        string
+	VolumeName  string
+	FsType      string
+}
```
mechairoi: [IMHO] `type FilesystemInfo` is only used by `util/windows/filesystem`. I think it's better to move to `util/windows/filesystem.go`.

metrics/windows/disk.go
```diff
@@ -106,7 +106,7 @@ func (g *DiskGenerator) Generate() (metrics.Values, error) {
 	results := make(map[string]float64)
 	for _, v := range g.counters {
 		var value windows.PDH_FMT_COUNTERVALUE_ITEM_DOUBLE
-		r, _, err := windows.PdhGetFormattedCounterValue.Call(uintptr(v.Counter), windows.PDH_FMT_DOUBLE, uintptr(0), uintptr(unsafe.Pointer(&value)))
+		r, _, err := windows.PdhGetFormattedCounterValue.Call(uintptr(v.Counter), windows.PDH_FMT_DOUBLE, uintptr(0), uintptr(unsafe.Pointer(&value.FmtValue)))
```
mechairoi: It looks like `PDH_FMT_COUNTERVALUE_ITEM_DOUBLE` is not used.
I think it is better to use `var value PDH_FMT_COUNTERVALUE_DOUBLE` and to get rid of all `PDH_FMT_COUNTERVALUE_ITEM_DOUBLE`.

metrics/windows/interface.go
```diff
@@ -53,6 +53,7 @@ func NewInterfaceGenerator(interval time.Duration) (*InterfaceGenerator, error)
 				name := windows.BytePtrToString(&ai.Description[0])
 				name = strings.Replace(name, "(", "[", -1)
 				name = strings.Replace(name, ")", "]", -1)
+				name = strings.Replace(name, "#", "_", -1)
```
mechairoi: :expressionless:

mattn:
LGTM, Thanks for correcting.

mechairoi:
LGTM, too.

util/util_windows.go
daiksy: I think so too!

metrics/windows/disk.go
daiksy: You 're right!

mechairoi:
LGTM :space_invader:

PR #64, opened 2015-02-06
Enhance windows installer

mechairoi:
- Remove installed files when uninstall mackerel-agent.
- Generate config file only if clean installation.
- Auto fill version into the wxs file from git tag.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@
 /build
 /id
 /pid
+/wix/mackerel-agent.wxs
+/wix/mackerel-agent.wixobj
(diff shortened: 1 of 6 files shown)


daiksy:
Where is the code of  `Remove installed files when uninstall mackerel-agent.` ?

mechairoi:
> Where is the code of  `Remove installed files when uninstall mackerel-agent.` ?

Probably it is default behavior of `<File>`.

daiksy:
thank you! LGTM!!

PR #65, opened 2015-02-11
add collector for ec2 metadata

stanaka:
EC2 has several metadata, such as instance-type, which are very useful for operators.
This adds a collector for EC2 metadata.

### DIFF
--- a/command/command_linux.go
+++ b/command/command_linux.go
@@ -9,12 +9,14 @@ import (
 )
 
 func specGenerators() []spec.Generator {
+	instanceGenerator, _ := specLinux.NewInstanceGenerator("")
 	return []spec.Generator{
 		&specLinux.KernelGenerator{},
 		&specLinux.CPUGenerator{},
 		&specLinux.MemoryGenerator{},
 		&specLinux.BlockDeviceGenerator{},
 		&specLinux.FilesystemGenerator{},
+		instanceGenerator,
 	}
 }
 
--- a/command/command_linux.go
+++ b/command/command_linux.go
@@ -16,11 +16,11 @@ func specGenerators() []spec.Generator {
 		&specLinux.BlockDeviceGenerator{},
 		&specLinux.FilesystemGenerator{},
 	}
-	instanceGenerator, err := specLinux.NewInstanceGenerator("")
+	cloudGenerator, err := specLinux.NewCloudGenerator("")
 	if err != nil {
-		logger.Errorf("Failed to create instanceGenerator: %s", err.Error())
+		logger.Errorf("Failed to create cloudGenerator: %s", err.Error())
 	} else {
-		specs = append(specs, instanceGenerator)
+		specs = append(specs, cloudGenerator)
 	}
 	return specs
 }
--- a/command/command_linux.go
+++ b/command/command_linux.go
@@ -20,7 +20,7 @@ func specGenerators() []spec.Generator {
 	if err != nil {
 		logger.Errorf("Failed to create instanceGenerator: %s", err.Error())
 	} else {
-		append(specs, instanceGenerator)
+		specs = append(specs, instanceGenerator)
 	}
 	return specs
 }
--- a/command/command_linux.go
+++ b/command/command_linux.go
@@ -9,15 +9,20 @@ import (
 )
 
 func specGenerators() []spec.Generator {
-	instanceGenerator, _ := specLinux.NewInstanceGenerator("")
-	return []spec.Generator{
+	specs := []spec.Generator{
 		&specLinux.KernelGenerator{},
 		&specLinux.CPUGenerator{},
 		&specLinux.MemoryGenerator{},
 		&specLinux.BlockDeviceGenerator{},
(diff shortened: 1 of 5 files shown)


spec/linux/instance.go
```diff
@@ -0,0 +1,87 @@
...
+	"time"
+)
+
+// EC2: http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/AESDG-chapter-instancedata.html
+// GCE: https://developers.google.com/compute/docs/metadata
+// DigitalOcean: https://developers.digitalocean.com/metadata/
+
+// InstanceGenerator XXX
+type InstanceGenerator struct {
+	BaseURL *url.URL
```
mechairoi: [IMHO] `BaseUrl` does not need to be published.

mechairoi:
LGTM :sun_with_face:

spec/linux/instance.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// Key XXX
```
motemen: We should write precisely for newly introduced comments.

command/command_linux.go
```diff
@@ -9,12 +9,14 @@ import (
 )
 
 func specGenerators() []spec.Generator {
+	instanceGenerator, _ := specLinux.NewInstanceGenerator("")
```
motemen: I don't think it's good to discard errors. Let's not include the generator to the result if any error is returned.

spec/linux/instance.go
```diff
@@ -0,0 +1,87 @@
...
+		}
+		defer resp.Body.Close()
+		if resp.StatusCode == 200 {
+			body, err := ioutil.ReadAll(resp.Body)
+			if err != nil {
+				instanceLogger.Errorf("Results of requesting metadata cannot be read.")
+				break
+			}
+			metadata[key] = string(body)
+			instanceLogger.Infof("results %s:%s", key, string(body))
```
motemen: This log may be a debug log so how about using Debugf?

motemen:
And one comment:

- When it became clear that the host it not on any cloud, shouldn't the generator stop accessing to 169.254.169.254?

stanaka:
The timeout of accessing 169.254.169.254 is set to 1 sec and if a timeout occurs, return statement called. So succeeding requests do not processed.
Of course a configuration which disable this generator may be useful. (dd-agent does so)

motemen:
I see, agreed with future configuration

spec/linux/instance.go
stanaka: I added some comments

command/command_linux.go
stanaka: Now errors are handled.

spec/linux/instance.go
stanaka: make it Debugf

stanaka:
I pushed some commits for review comments.
Please re-review them.

spec/linux/cloud.go
```diff
@@ -0,0 +1,91 @@
...
+				cloudLogger.Errorf("Results of requesting metadata cannot be read.")
+				break
+			}
+			metadata[key] = string(body)
+			cloudLogger.Debugf("results %s:%s", key, string(body))
+		}
+	}
+
+	results := make(map[string]interface{})
+	results["provider"] = "ec2"
```
motemen: Is it okay to call this just "ec2", not "aws ec2" for example?

spec/linux/cloud.go
```diff
@@ -0,0 +1,91 @@
...
+		resp, err := client.Get(g.baseURL.String() + "/" + key)
+		if err != nil {
+			cloudLogger.Infof("This host may not be running on EC2. Error while reading '%s'", key)
+			return nil, nil
+		}
+		defer resp.Body.Close()
+		if resp.StatusCode == 200 {
+			body, err := ioutil.ReadAll(resp.Body)
+			if err != nil {
+				cloudLogger.Errorf("Results of requesting metadata cannot be read.")
```
motemen: Including `err` in the log would be better.

spec/linux/cloud.go
```diff
@@ -0,0 +1,91 @@
...
+	metadata := make(map[string]string)
+
+	for _, key := range metadataKeys {
+		resp, err := client.Get(g.baseURL.String() + "/" + key)
+		if err != nil {
+			cloudLogger.Infof("This host may not be running on EC2. Error while reading '%s'", key)
+			return nil, nil
+		}
+		defer resp.Body.Close()
+		if resp.StatusCode == 200 {
```
motemen: I'd like to log if resp.StatusCode is not 200.

motemen:
LGTM

spec/linux/cloud.go
stanaka: I think "aws ec2" is redundant because the abbreviation like "ec2" is very popular.
but just "aws" which covers wider services amazon provides may be better than "ec2".

motemen: I see :blush:

PR #72, opened 2015-03-08
Run once and output results to stdout

stanaka:
Add `-once` option that mackerel-agent runs once and outputs results to stdout.
This option is useful for developing and debugging mackerel-agent itself and plugins.

Example of output is as follows.
```json
# build/mackerel-agent -once -conf=mackerel-agent.conf | jq .
2015/03/08 12:46:46 INFO main Starting mackerel-agent version:0.14.3, rev:85bd358
2015/03/08 12:46:46 INFO command Collecting metrics may take one minutes.
{
  "host": {
    "interfaces": null,
    "meta": {
      "agent-name": "mackerel-agent/0.14.3 (Revision 85bd358)",
      "agent-revision": "85bd358",
      "agent-version": "0.14.3",
      "cpu": [
        {
          "model_name": "Intel(R) Core(TM) i7-3667U CPU @ 2.00GHz\n"
        }
      ],
      ...
}
```

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -90,20 +90,9 @@ func (agent *Agent) CollectGraphDefsOfPlugins() []mackerel.CreateGraphDefsPayloa
 	return payloads
 }
 
-
 // InitPluginGenerators XXX
 func (agent *Agent) InitPluginGenerators(api *mackerel.API) {
-	payloads := []mackerel.CreateGraphDefsPayload{}
-
-	for _, g := range agent.PluginGenerators {
-		p, err := g.PrepareGraphDefs()
-		if err != nil {
-			logger.Debugf("Failed to fetch meta information from plugin %s (non critical); seems that this plugin does not have meta information: %s", g, err)
-		}
-		if p != nil {
-			payloads = append(payloads, p...)
-		}
-	}
+	payloads := agent.CollectGraphDefsOfPlugins()
 
 	if len(payloads) > 0 {
 		err := api.CreateGraphDefs(payloads)
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -20,7 +20,7 @@ type MetricsResult struct {
 	Values  metrics.Values
 }
 
-func (agent *Agent) collectMetrics(collectedTime time.Time) *MetricsResult {
+func (agent *Agent) CollectMetrics(collectedTime time.Time) *MetricsResult {
 	generators := agent.MetricsGenerators
 	for _, g := range agent.PluginGenerators {
 		generators = append(generators, g)
@@ -63,7 +63,7 @@ func (agent *Agent) Watch() chan *MetricsResult {
 			ti := tickedTime
 			sem <- 1
 			go func() {
-				metricsResult <- agent.collectMetrics(ti)
+				metricsResult <- agent.CollectMetrics(ti)
 				<-sem
 			}()
 		}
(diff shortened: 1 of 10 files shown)


agent/agent.go
```diff
@@ -72,12 +73,30 @@ func (agent *Agent) Watch() chan *MetricsResult {
...
+		if err != nil {
+			logger.Debugf("Failed to fetch meta information from plugin %s (non critical); seems that this plugin does not have meta information: %s", g, err)
+		}
+		if p != nil {
+			payloads = append(payloads, p...)
+		}
+	}
+
+	return payloads
+}
```
motemen: Can InitPluginGenerators() be modified to use this function?

command/command.go
```diff
@@ -346,6 +346,37 @@ func Prepare(conf *config.Config) (*mackerel.API, *mackerel.Host, error) {
...
+	metrics := ag.CollectMetrics(time.Now())
+	payload := map[string]interface{}{
+		"host": map[string]interface{}{
+			"name":          hostname,
+			"meta":          meta,
+			"interfaces":    interfaces,
+			"roleFullnames": conf.Roles,
+		},
+		"metrics": metrics,
+	}
```
motemen: This data structure reflects what `mackerel.UpdateHost()` sents to the API server, so this part should be moved to another function which would also be used by UpdateHost, not duplicating code.

command/command.go
```diff
@@ -346,6 +346,37 @@ func Prepare(conf *config.Config) (*mackerel.API, *mackerel.Host, error) {
...
+	json, err := json.Marshal(payload)
+	if err != nil {
+		logger.Warningf("Error while marshaling graphdefs: graphdefs = %s.", graphdefs)
```
motemen: It would be better to include `err` to the log

main.go
```diff
@@ -121,6 +127,8 @@ func resolveConfig() (*config.Config, bool) {
 			conf.Verbose = verbose
 		case "role":
 			conf.Roles = roleFullnames
+		case "once":
+			conf.OutputOnce = *outputOnce
```
motemen: OutputOnce is not an agent configuration but a CLI program behavior like printVersion (for `-version`), so we should not include OutputOnce to Config. I think it would be good to extend the return value of resolveConfig to pass `-once` flags, instead of appending it to conf.

command/command.go
```diff
@@ -346,6 +346,37 @@ func Prepare(conf *config.Config) (*mackerel.API, *mackerel.Host, error) {
...
+	} else {
+		fmt.Println(string(json))
```
motemen: [IMO] We may use `json.Indent()` to prettify the JSON

motemen:
Added some comments :sunglasses:

stanaka:
I added some commits. re-review plz!!

main.go
```diff
@@ -41,12 +41,17 @@ func (r *roleFullnamesFlag) Set(input string) error {
 	return nil
 }
 
+type otherOptions struct {
+	PrintVersion bool
+	RunOnce      bool
```
motemen: These fields could be lowerCamelCased?

main.go
```diff
@@ -97,7 +109,13 @@ func resolveConfig() (*config.Config, bool) {
...
 
 	if *printVersion {
-		return nil, true
+		otherOptions.PrintVersion = true
+		return nil, otherOptions
+	}
+
+	if *runOnce {
+		otherOptions.RunOnce = true
+		return nil, otherOptions
```
motemen: Returning nil as *config.Config here causes nil pointer panic on the caller side. https://github.com/mackerelio/mackerel-agent/blob/b54cbc5/main.go#L59

motemen:
Mostly looks good, but some generators require non-nil Config, so I think we must create one in resolveConfig() even if `-once` is provided.

main.go
stanaka: oops

PR #75, opened 2015-03-31
output GOOS GOARCH runtime.Version() when -version option is specified

Songmu:
% mackerel-agent -version
    mackerel-agent version 0.14.3 (rev a8702fe) [darwin amd64 go1.4]


mechairoi:
LGTM!

PR #80, opened 2015-04-14
"Check" functionality

motemen:
:speak_no_evil: 

rough specs
-----------

- mackerel-agent invokes commands specified in the configuration file periodically
- The agent sends reports from the result of the commands to Mackerel API
  - then Mackerel opens alerts depending the reports
- The result of command invocation is treated as below:
  - exit status: 0=OK 1=Warning 2=Critical (3=Unknown)
  - stdout is treated as message (to be shown in the alert page and notifications)

code changes
------------
- now command has other loops where
  - one for reporting
  - one for each commands

### DIFF
--- a/command/command.go
+++ b/command/command.go
@@ -231,7 +231,9 @@ func loop(c *context, termCh chan struct{}) int {
 					}
 				}
 
-				logger.Debugf("reports: %#v", reports)
+				for i, report := range reports {
+					logger.Debugf("reports[%d]: %#v", i, report)
+				}
 
 				if len(reports) == 0 {
 					continue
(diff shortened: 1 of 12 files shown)


command/command.go
```diff
@@ -304,6 +318,113 @@ func enqueueLoop(c *context, postQueue chan *postValue, quit chan struct{}) {
 	}
 }
 
+// runCheckersLoop generates "checker" goroutines
+// which run for each checker commands and one for HTTP POSTing
+// the reports to Mackerel API.
+func runCheckersLoop(c *context, termCheckerCh chan struct{}, quit chan struct{}) {
```
mechairoi: [IMO] `termCheckerCh <-chan struct{}, quit <-chan struct{}` is better.

motemen: Reasonable :ok_woman:

PR #84, opened 2015-05-06
collect memory metrics of osx

Songmu:
Retrieve memroy metrics from `vm_stat` command.
These values are almost same as Activity Monitor.

ref. https://support.apple.com/en-us/HT201464#memory

### DIFF
--- a/metrics/darwin/memory.go
+++ b/metrics/darwin/memory.go
@@ -104,10 +104,10 @@ func (g *MemoryGenerator) Generate() (metrics.Values, error) {
 	} else {
 		if matches := swapReg.FindStringSubmatch(string(outBytes)); matches != nil {
 			t, _ := strconv.ParseFloat(matches[1], 64)
-			u, _ := strconv.ParseFloat(matches[2], 64)
+			// swap_used are calculated at server, so don't send it
+			// u, _ := strconv.ParseFloat(matches[2], 64)
 			f, _ := strconv.ParseFloat(matches[3], 64)
 			ret["memory.swap_total"] = t * 1024 * 1024
-			ret["memory.swap_used"] = u * 1024 * 1024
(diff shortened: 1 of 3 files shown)


metrics/darwin/memory.go
```diff
@@ -0,0 +1,115 @@
...
+*/
+
+/* sysctl vm.swapusage sample
+% sysctl vm.swapusage
+vm.swapusage: total = 1024.00M  used = 2.50M  free = 1021.50M  (encrypted)
+*/
+
+var memoryLogger = logging.GetLogger("metrics.memory")
+var statReg = regexp.MustCompile(`^(.+):\s+([0-9]+)\.$`)
+var swapReg = regexp.MustCompile(`([0-9]+(?:\.[0-9]+))?M[^0-9]*([0-9]+(?:\.[0-9]+)?)M[^0-9]*([0-9]+(?:\.[0-9]+)?)M`)
```
mechairoi: Swap the second `?` for `)` ?

metrics/darwin/memory.go
```diff
@@ -0,0 +1,115 @@
...
+		memoryLogger.Errorf("Failed (skip swap metrics): %s", err)
+	} else {
+		if matches := swapReg.FindStringSubmatch(string(outBytes)); matches != nil {
+			t, _ := strconv.ParseFloat(matches[1], 64)
+			// swap_used are calculated at server, so don't send it
+			// u, _ := strconv.ParseFloat(matches[2], 64)
+			f, _ := strconv.ParseFloat(matches[3], 64)
+			ret["memory.swap_total"] = t * 1024 * 1024
+			ret["memory.swap_free"] = f * 1024 * 1024
+		}
```
mechairoi: [nit] I think it is better to `Errorf` also when match is failed.

PR #85, opened 2015-05-06
Code sharing around dfValues

Songmu:
util/linux/filesystem.go, util/freebsd/filesystem.go and util/darwin/filesystem.go are close to the same. So commonalize them.


mechairoi:
LGTM!

PR #83, opened 2015-04-24
Send plugin.check._name_s list on `updateHost`

mechairoi:

PR #86, opened 2015-05-07
Merge #83 into master branch. (Send plugin.check._name_s list on `updateHost`)

mechairoi:


Songmu:
Merge #83 into master branch. (Send plugin.check._name_s list on `updateHost`)

PR #92, opened 2015-05-13
Set `displayName` via agent

Sixeight:
Host has "Display Name" to manage host using any name.
You can set "Display Name" via Web UI but you cannot set via agent.

This PR proposes new `displayName` setting.
And send the setting with other host specs when agent has started.

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -11,6 +11,7 @@ import (
 
 var sampleConfig = `
 apikey = "abcde"
+displayname = "fghij"
 
 [connection]
 post_metrics_retry_delay_seconds = 600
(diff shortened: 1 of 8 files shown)


config/config_test.go
```diff
@@ -11,6 +11,7 @@ import (
 
 var sampleConfig = `
 apikey = "abcde"
+displayname = "fghij"
```
Songmu: [imo] `display_name` is better.

Songmu:
@Sixeight  LGTM!

Could you review #93, before merging this p-r?

Sixeight:
@Songmu thank you! ok, I'll try to review your p-r!

PR #95, opened 2015-05-30
Refactor logging

Songmu:
- define type logging.level
- use `stringer` and `go generate` for maintainability
- output filename and lines when the log-level below DEBUG


Sixeight:
LGTM!

PR #96, opened 2015-05-31
Enhance api testings

Songmu:


Songmu:
add only testings. so merge it.

PR #98, opened 2015-06-01
remove lockfile feature in initd

Songmu:
The lockfile doesn't needed. It is enough only using pidfile.


mattn:
BTW, why remove this?

PR #100, opened 2015-06-03
Configurable host status on start/stop agent

Songmu:
Specify host status on start/stop agent in configuration as follows.

```
[host_status]
on_start = "working"
on_stop  = "poweroff"
```

### DIFF
--- a/command/command.go
+++ b/command/command.go
@@ -543,7 +543,12 @@ func Run(conf *config.Config, api *mackerel.API, host *mackerel.Host, termCh cha
 		api:  api,
 	}
 
-	return loop(c, termCh)
+	exitCode := loop(c, termCh)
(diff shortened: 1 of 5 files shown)


command/command.go
```diff
@@ -536,7 +543,12 @@ func Run(conf *config.Config, api *mackerel.API, host *mackerel.Host, termCh cha
 		api:  api,
 	}
 
-	return loop(c, termCh)
+	exitCode := loop(c, termCh)
+	if exitCode == 0 && conf.HostStatus.Stop != "" {
+		// TOOD error handling. supoprt retire
+		api.UpdateHostStatus(host.ID, conf.HostStatus.Stop)
```
Sixeight: I think we need some output at least when failed update host status.

PR #101, opened 2015-06-05
Add an agent memory usage metrics generator for diagnostic use

hakobe:
To avoid memory leaks, investigating memory usage of the agent should be available.

This PR  includes a metrics generator that collects metrics of memory usage of the agent. Those metrics are sent as custom metrics. This functionality is can be turned on by putting `diagnostic_mode` configuration to the agent config file.


Songmu:
LGTM!

PR #102, opened 2015-06-05
Add mkr to deb/rpm package

Sixeight:
I want to include mkr into deb/rpm package.
Because some operation with mkr on server that is installed mackerel-agent. 
(check status, retire and so on)


Songmu:
LGTM!

Sixeight:
Thank you!

PR #109, opened 2015-07-03
[WORKAROUND] downgrade golang version for windows

Sixeight:
We have a some problem with go 1.4.2 on windows agent.
Thus, I propose temporary fix that using go 1.3.3 for windows build.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,8 +7,8 @@ install:
   - echo %Path%
   - choco install ruby
   - rd c:\go /s /q
-  - appveyor DownloadFile https://storage.googleapis.com/golang/go1.4.2.windows-amd64.zip
-  - 7z x go1.4.2.windows-amd64.zip -oC:\ >nul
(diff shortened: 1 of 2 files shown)


mattn:
What problems do you have?

Sixeight:
We have a problem which seems to be same as this issue: https://github.com/golang/go/issues/10026
Do you have any information about the problem?

Songmu:
The mackerel-agent built with go 1.4 often crashes suddenly in windows environment, and built with go 1.3 doesn't.

So this workaround is reasonable change.

:+1:

mattn:
I'm not having objection about this. However, if you get the stacktrace on panic, please let me see.

Songmu:
paniced on Windows Server 2008 on EC2.

https://gist.github.com/Songmu/a4136fc99fc67dfeb61d

mattn:
Thank you for taking time to do it. But unfortunately, I have no idea to fix this issue.

Sixeight:
Thank you for your help!

PR #119, opened 2015-07-15
allow using spaces as delimiter for split custom metric values

### LINKED ISSUE #118: Custom metrics input values should be split with space not tab
According to the [Mackerel document](http://help.mackerel.io/entry/advanced/custom-metrics), custom metrics plugins should output tab-separated values as follows;

```
{metric name}\t{metric value}\t{epoch seconds}
```

but Sensu plugins using Sensu::Plugin::Metric::CLI::Graphite output space-separated values;

https://github.com/sensu-plugins/sensu-plugin/blob/master/lib/sensu-plugin/metric/cli.rb#L26

so if you say the format is "compatible with sensu", the parser should split values with spaces not tabs or both of them.

https://github.com/mackerelio/mackerel-agent/blob/master/metrics/plugin.go#L231

Songmu:
fix #118

### DIFF
--- a/metrics/plugin.go
+++ b/metrics/plugin.go
@@ -212,6 +212,8 @@ func (g *pluginGenerator) makeCreateGraphDefsPayload() []mackerel.CreateGraphDef
 	return payloads
 }
 
+var delimReg = regexp.MustCompile(`[\s\t]+`)
+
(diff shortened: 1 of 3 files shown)


Sixeight:
LGTM!

PR #134, opened 2015-08-05
use C struct for accessing Windows APIs

stanaka:
Using a pointer of Go struct as an argument of Windows API causes panic. To fix this, C struct should be used.

This pull request provides a helper method to retrieve a value with C struct of Windows API, and changes to use it, not to use methods with Go struct.

ref. https://github.com/lxn/win/issues/12


Songmu:
Great!

mattn:
:+1:

PR #135, opened 2015-08-11
Fix bug that checks is not removed when no checks.

Sixeight:
## Problem

When last check is removed, agent does not send `checks`.

## Solution

Send empty list as `checks`.


Songmu:
LGTM!

Sixeight:
thanks!

PR #144, opened 2015-09-13
Correct sample nginx comment.

kamatama41:
At the mackerel-plugin-nginx, the default value of `port` is [8080](https://github.com/mackerelio/mackerel-agent-plugins/blob/master/mackerel-plugin-nginx/nginx.go#L173).


itchyny:
LGTM

PR #145, opened 2015-09-14
Release version 0.23.0

itchyny:
- send check monitor report to server when check script failed even if the monitor result is not changed #143
- Correct sample nginx comment. #144


Songmu:
LGTM

itchyny:
Thank you.

PR #152, opened 2015-10-17
/proc/cpuinfo parser for old ARM Linux kernels

hanazuki:
Linux kernels prior to 3.8 has a different format of /proc/cpuinfo for ARM processors
than that for x86 ones.

In those old kernels the format depends on the CONFIG_SMP kernel configuration.
With the configuration enabled, /proc/cpuinfo has a `Processor: ` line with the CPU's
modle name and a `processor: %d` line for each CPU cores.

With the configuration disabled (which implies that the kernel handles only one CPU),
/proc/cpuinfo has a `Processor: ` line with the CPU's model name and no `processor: ` line.

ref: https://github.com/torvalds/linux/commit/b4b8f770eb10a1bccaf8aa0ec1956e2dd7ed1e0a

### DIFF
--- a/spec/linux/cpu.go
+++ b/spec/linux/cpu.go
@@ -4,6 +4,7 @@ package linux
 
 import (
 	"bufio"
+	"io"
 	"os"
 	"regexp"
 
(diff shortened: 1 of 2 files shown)


Songmu:
Thank you!

PR #169, opened 2015-12-01
Configtest

fujiwara:
related #94.

- Added `-configtest` option to mackerel-agent.
- Added argument `configtest` to init scripts.


Songmu:
Great! Thanks!

PR #172, opened 2015-12-02
fix exit status of debian init script.

fujiwara:
An init script should exit with status non 0 when failed on any action.


Songmu:
:pray:

PR #187, opened 2016-02-01
fix the code of extending the process's environment

itchyny:
The original code seems to append `LANG=C` to the environment variables, but cmd.Env is nil until it is set. This pull request fixes the code; use `os.Environ()` to obtain the current process's environment.


Songmu:
LGTM

itchyny:
Thank you.

PR #192, opened 2016-02-10
fix the exit status on stopping the agent in the init script of debian

itchyny:
This pull request fixes the exit status of `/etc/init.d/mackerel-agent stop` on debian. With the current script, the `stop` command exits with 1 in most case; when `$AUTO_RETIREMENT` is not set.

ref: #172


Songmu:
LGTM

itchyny:
Thank you.

PR #197, opened 2016-02-25
Report checker execution timeout as Unknown status

hanazuki:
In the current implementation timeouts of check plugins are only logged to the agent's log but not reported to the server.
Since a checker command timeouts due to unusual conditions such as a heavy load on the machine, unstable networking or other misconfigurations, I suppose the events shall be immediately reported as incidents.


Songmu:
Good fix! Thank you!

Songmu:
I implemented test for this p-r based on your code at 64269c. Please check it.

hanazuki:
Thank you, Songmu-san!

PR #208, opened 2016-03-15
Refactor net interface (multi ip support and bug fix)

Songmu:
- define spec.NetInterface
- define spec.InterfaceGenerator and change command/interfaceGenerator using it
- multiple ip support in windows and darwin
- distinguish ipv4 and ipv6 in spec/windows/interface


mechairoi:
LGTM :+1: :+1: :+1:

PR #209, opened 2016-03-16
Stop to fetch flags of cpu in spec/linux/cpu

Songmu:
Stop to fetch flags of cpu in spec/linux/cpu. It's too noisy


itchyny:
LGTM!

PR #213, opened 2016-03-25
use GOARCH=amd64 for now

Songmu:


udzura:
Dropping i386 support?

stanaka:
Currently we have some trouble with i386 binary on Amazon Linux 2016.03, so this is inteded to be a temporary fix.
I think it is related with this issue https://github.com/golang/go/issues/14795 .
We're going to change packagings of mackerel-agent for supporting both amd64 and i386.

PR #226, opened 2016-04-28
output plugin stderr to log

Songmu:


metrics/plugin.go
```diff
@@ -221,8 +221,11 @@ func (g *pluginGenerator) collectValues() (Values, error) {
 	os.Setenv(pluginConfigurationEnvName, "")
 	stdout, stderr, _, err := util.RunCommand(command)
 
+	if stderr != "" {
+		pluginLogger.Infof("command %q printed out on STDERR: %q", stderr)
```
stanaka: `"Command %q outputted to STDERR: %q"`may be better.

stanaka:
Just one comment about message. 
Regarding others, LTGM

PR #228, opened 2016-05-02
Recover from panic while processing generators

stanaka:
Generator may panic in some case, such as running mackerel-agent in new/old/modified environments.
With this patch, mackerel-agent recovers from panic of processing generators and continues to work.


Songmu:
LGTM

PR #235, opened 2016-05-20
remove regexp from spec/linux/cpu

Songmu:
see https://github.com/mackerelio/mackerel-agent/pull/235/files?w=1 for review.


itchyny:
LGTM!

PR #243, opened 2016-06-10
Don't stop mackerel-agent process on upgrading by debian package

karupanerura:
Current `prerm` script stop mackerel-agent daemon. And this script is run on upgrading.
But, I don't happy to stop the mackerel-agent daemon on upgrading.
So I suggest to stop the mackerel-agent daemon on removing only.


karupanerura:
SEE ALSO: https://www.debian.org/doc/debian-policy/ch-maintainerscripts.html

Songmu:
Great! 🙆

PR #245, opened 2016-06-13
Added plugin check interval option

karupanerura:
I suggest to add `check_interval` option to checker plugin configuration.
I have to run a too heavy checker script...

### DIFF
--- a/checks/checker.go
+++ b/checks/checker.go
@@ -98,10 +98,10 @@ func (c Checker) Check() (*Report, error) {
 func (c Checker) Interval() time.Duration {
 	if c.Config.ExecutionInterval != nil {
 		interval := time.Duration(*c.Config.ExecutionInterval) * time.Minute
-		if interval < 1*time.Second {
-			interval = 1 * time.Second
-		} else if interval > 60*time.Second {
-			interval = 60 * time.Second
+		if interval < 1*time.Minute {
+			interval = 1 * time.Minute
+		} else if interval > 60*time.Minute {
+			interval = 60 * time.Minute
 		}
 		return interval
 	}
--- a/checks/checker.go
+++ b/checks/checker.go
@@ -25,7 +25,7 @@ const (
 	StatusUnknown   Status = "UNKNOWN"
 )
 
-const defaultExecutionInterval = 1 * time.Minute
+const defaultCheckInterval = 1 * time.Minute
 
 var exitCodeToStatus = map[int]Status{
 	0: StatusOK,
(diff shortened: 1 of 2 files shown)


Songmu:
It's a nice pull request!

checks/checker.go
```diff
@@ -94,5 +96,8 @@ func (c Checker) Check() (*Report, error) {
 // Interval is the interval where the command is invoked.
 // (Will be configurable in the future)
 func (c Checker) Interval() time.Duration {
```
Songmu: limit the range of return values.
It should be between 1 to 60 minutes.

If the configured value is out of range, 
truncate it to minimum or maximum value in the range
with warning logs.

config/config.go
```diff
@@ -68,6 +68,7 @@ type PluginConfigs map[string]PluginConfig
 type PluginConfig struct {
 	Command              string
 	NotificationInterval *int32  `toml:"notification_interval"`
+	ExecutionInterval    *int32  `toml:"execution_interval"`
```
Songmu: ExecutionInterval doesn't work in metric plugins.
It's OK. I think it is preferred behaviour.
Please comment it.

Songmu:
I pointed out the two points. Please fix them.
Thanks!

karupanerura:
Thank you for your review. I'll fix it in near days.

Songmu:
OK. Please `gofmt` at last!

karupanerura:
oops! thank you for your notice.

karupanerura:
done!

checks/checker.go
```diff
@@ -94,5 +96,14 @@ func (c Checker) Check() (*Report, error) {
...
-	return 1 * time.Minute
+	if c.Config.ExecutionInterval != nil {
+		interval := time.Duration(*c.Config.ExecutionInterval) * time.Minute
+		if interval < 1*time.Second {
+			interval = 1 * time.Second
```
itchyny: s/Second/Minute/

karupanerura:
🙇

checks/checker.go
karupanerura: 🙇

config/config.go
```diff
@@ -64,10 +64,11 @@ type Config struct {
 type PluginConfigs map[string]PluginConfig
 
 // PluginConfig represents a section of [plugin.*].
-// `MaxCheckAttempts` option is used with check monitoring plugins. Custom metrics plugins ignore this option.
+// `MaxCheckAttempts`, `NotificationInterval`, and `ExecutionInterval` option are used with check monitoring plugins. Custom metrics plugins ignore this option.
```
itchyny: `options are` ... `ignore these options`

karupanerura: oops! 🙇

itchyny:
Looks Good to Me!

Issue #77, opened 2015-04-04
Specify username for executing a plugin


stanaka:
To change username, `sudo` must be used as work-around like:
```
command="sudo -u xxxx ..."
```

Specifying username is as follows.
```
[plugin.metrics.vmstat]
command = "ruby /etc/sensu/plugins/system/vmstat-metrics.rb"
username = nobody
```

Songmu:
resolved at #250

PR #250, opened 2016-06-29
Add user for executing a plugin

y-kuno:
add `user_name` option to plugin configuration

### DIFF
--- a/checks/checker.go
+++ b/checks/checker.go
@@ -65,11 +65,9 @@ func (c Checker) String() string {
 func (c Checker) Check() (*Report, error) {
 	now := time.Now()
 
-	var command string
+	command := c.Config.Command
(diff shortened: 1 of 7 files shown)


Songmu:
ref #77 

It is a nice pull request! We have wanted to do it for a long time.

- I think just `user` is enough instead of `user_name`.
- `sudo` doesn't work on windows.
- Could you change the interface of `utils.RunCommand()`?
  - ex. `RunCommand(command, user string) ...`
  - `sudo -u $user /bin/sh -c "$command"` is safer than `/bin/sh -c "sudo -u $user $command"`
  - On windows, please ignore `user` argument with warning log.

y-kuno:
Thank you for your review. I'll fix it in near days.

Songmu:
Good Job!

PR #255, opened 2016-07-15
Fill the customIdentifier in EC2

itchyny:
With this pull request, mackerel-agent on cloud platforms can identify the host by customIdentifier. Currently, EC2 is only supported.

Firstly, the agent on EC2 decides the customIdentifier based on the instance-id. Before creating a new host, the agent tries to find a host by the customIdentifier (if found, the host is possibly created by the AWS integration). On upgrading the version of mackerel-agent, it updates the host information with customIdentifier filled.

### DIFF
--- a/command/command.go
+++ b/command/command.go
@@ -509,12 +509,13 @@ func (c *Context) UpdateHostSpecs() {
 	}
 
 	err = c.API.UpdateHost(c.Host.ID, mackerel.HostSpec{
-		Name:          hostname,
-		Meta:          meta,
-		Interfaces:    interfaces,
-		RoleFullnames: c.Config.Roles,
-		Checks:        c.Config.CheckNames(),
-		DisplayName:   c.Config.DisplayName,
+		Name:             hostname,
+		Meta:             meta,
+		Interfaces:       interfaces,
+		RoleFullnames:    c.Config.Roles,
+		Checks:           c.Config.CheckNames(),
+		DisplayName:      c.Config.DisplayName,
+		CustomIdentifier: customIdentifier,
 	})
 
 	if err != nil {
@@ -581,12 +582,13 @@ func runOncePayload(conf *config.Config) ([]mackerel.CreateGraphDefsPayload, *ma
 	graphdefs := ag.CollectGraphDefsOfPlugins()
 	metrics := ag.CollectMetrics(time.Now())
 	return graphdefs, &mackerel.HostSpec{
-		Name:          hostname,
-		Meta:          meta,
-		Interfaces:    interfaces,
-		RoleFullnames: conf.Roles,
-		Checks:        conf.CheckNames(),
-		DisplayName:   conf.DisplayName,
+		Name:             hostname,
+		Meta:             meta,
+		Interfaces:       interfaces,
+		RoleFullnames:    conf.Roles,
+		Checks:           conf.CheckNames(),
+		DisplayName:      conf.DisplayName,
+		CustomIdentifier: customIdentifier,
 	}, metrics, nil
 }
 
--- a/command/command.go
+++ b/command/command.go
@@ -54,7 +54,6 @@ func prepareHost(conf *config.Config, api *mackerel.API) (*mackerel.Host, error)
 
 	var result *mackerel.Host
 	if hostID, err := conf.LoadHostID(); err != nil { // create
-		logger.Debugf("Registering new host on mackerel...")
 
 		if customIdentifier != "" {
 			doRetry(func() error {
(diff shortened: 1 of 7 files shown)


command/command.go
```diff
@@ -56,13 +56,32 @@ func prepareHost(conf *config.Config, api *mackerel.API) (*mackerel.Host, error)
...
+				})
+				return filterErrorForRetry(lastErr)
+			})
+
+			if lastErr != nil {
+				return nil, fmt.Errorf("Failed to register this host: %s", lastErr.Error())
+			}
 		}
 
 		doRetry(func() error {
```
haya14busa: I guess we can move this block into  `if result == nil { ...} `

We have no need to retry to find host by id.

haya14busa: or, does it mean to check the host found by custom identifier has correct hostId? then, I want some comment.

mackerel/host.go
```diff
@@ -10,10 +10,11 @@ type Host struct {
...
-	RoleFullnames []string               `json:"roleFullnames"`
-	Checks        []string               `json:"checks"`
-	DisplayName   string                 `json:"displayName,omitempty"`
+	Name             string                 `json:"name"`
+	Meta             map[string]interface{} `json:"meta"`
+	Interfaces       interface{}            `json:"interfaces"`
+	RoleFullnames    []string               `json:"roleFullnames"`
+	Checks           []string               `json:"checks"`
+	DisplayName      string                 `json:"displayName,omitempty"`
+	CustomIdentifier *string                `json:"customIdentifier,omitempty"`
```
haya14busa: [just a comment?] I don't quite understand  using both string pointer and omitempty.

I guess just using omitempty and string type may be enough.
If we use both string pointer and omitempty, we may path string value.

https://play.golang.org/p/OZb7TXdXO1

spec/cloud.go
```diff
@@ -153,6 +154,30 @@ func (g *EC2Generator) Generate() (interface{}, error) {
 	return results, nil
 }
 
+// SuggestCustomIdentifier suggests the identifier of the EC2 instance
+func (g *EC2Generator) SuggestCustomIdentifier() (*string, error) {
```
haya14busa: it's not so much complicated code, but adding some simple test is preferable, i think

mackerel/host.go
itchyny: Thanks, you've got a point. I really dislike the code regarding the empty string as nothing so I wanted to return nil when the custom identifier is not available. But I didn't notice the behaviour of `omitempty` you pointed out, `string` is surely preferable in this case.

command/command.go
itchyny: Good catch. I think I was too nervous...

command/command.go
```diff
@@ -56,21 +56,40 @@ func prepareHost(conf *config.Config, api *mackerel.API) (*mackerel.Host, error)
...
-
-		if lastErr != nil {
-			return nil, fmt.Errorf("Failed to register this host: %s", lastErr.Error())
+		if customIdentifier != "" {
+			doRetry(func() error {
+				result, lastErr = api.FindHostByCustomIdentifier(customIdentifier)
+				return filterErrorForRetry(lastErr)
+			})
+			if result != nil {
+				hostID = result.ID
```
haya14busa: now, you can remove this code.

command/command.go
```diff
@@ -56,21 +56,40 @@ func prepareHost(conf *config.Config, api *mackerel.API) (*mackerel.Host, error)
 	if hostID, err := conf.LoadHostID(); err != nil { // create
 		logger.Debugf("Registering new host on mackerel...")
```
haya14busa: maybe, this message should be after `if customIdentifier != "" {...}` block.

itchyny:
Thanks for the review.

yoheimuta:
[MEMO] We found a more (+1 min) delay to create a /var/lib/mackerel-agent/id file. This change has an impact for ec2 users who for example uses an auto scaling feature and mackerel apis. 

- Before (v0.32.2), the file is created and then cloud-init started. 
- Now (v0.33.0), cloud-init started and the file is created later.

```
2016/08/09 15:11:42 INFO <main> Starting mackerel-agent version:0.33.0, rev:f3a97ae, apibase:https://mackerel.io

# This process is retried for 60 seconds
## https://github.com/mackerelio/mackerel-agent/compare/v0.32.2...v0.33.0#diff-f2f57f6e81c1ea7ce925abd60673525cR59
2016/08/09 15:11:42 WARNING <command> No host was found for the custom identifier: i-d7597e48.ec2.amazonaws.com
...
2016/08/09 15:12:40 WARNING <command> No host was found for the custom identifier: i-d7597e48.ec2.amazonaws.com
2016/08/09 15:12:40 INFO <command> Start: apibase = https://mackerel.io, hostName = ip-10-0-1-207, hostID = 2LHbdNAA4yC
```

PR #258, opened 2016-08-10
Reduce retry count on finding a host by the custom identifier

itchyny:
#255

https://github.com/mackerelio/mackerel-agent/pull/255#issuecomment-238782597


Songmu:
LGTM

itchyny:
Thank you.

PR #260, opened 2016-08-11
suppress checker flooding when resuming from sleep mode

Songmu:
- care laptop sleep mode
- refactor
  - remove utils.Periodically
    - remove race condition when updating lastStatus and lastMessage
  - reduce goroutines
  - pointerize Checker


itchyny:
LGTM!

PR #262, opened 2016-08-12
commonalize spec.FilesystemGenerator around unix OSs

Songmu:


spec/filesystem.go
```diff
@@ -16,8 +13,6 @@ func (g *FilesystemGenerator) Key() string {
 	return "filesystem"
 }
 
-var logger = logging.GetLogger("spec.filesystem")
```
astj: [memo]
`logger` has been defined but not used

astj:
[memo]
Those `dfColumnSpecs` are all same

astj:
LGTM

PR #263, opened 2016-08-12
define type DfStat,	remove dfColumnSpecs and refactor

Songmu:
follow on #262


astj:
[memo]
Changes after #262: https://github.com/mackerelio/mackerel-agent/compare/1d3d98765ed9f4557d9e643ca4e5e3b227a90376...df

metrics/filesystem.go
```diff
@@ -14,38 +14,28 @@ type FilesystemGenerator struct {
...
 
-var dfColumnSpecs = []util.DfColumnSpec{
-	{Name: "size", IsInt: true},
-	{Name: "used", IsInt: true},
-}
-
 var sanitizerReg = regexp.MustCompile(`[^A-Za-z0-9_-]`)
 
 // Generate the metrics of filesystems
 func (g *FilesystemGenerator) Generate() (Values, error) {
```
astj: I found that tests for `Generate()` is so simple 😱
https://github.com/mackerelio/mackerel-agent/blob/master/metrics/filesystem_test.go

util/filesystem.go
```diff
@@ -18,6 +19,15 @@ import (
 	"github.com/mackerelio/mackerel-agent/logging"
 )
 
+type DfStat struct {
+	Name      string
+	Blocks    uint64
```
astj: [suggestion]
"Blocks" and "Capacity" are named "size" and "percent_used" in spec. Those mismatch are confusing, so how about using same names as spec?

astj:
Reviewed.

Songmu:
The field names of `dfStat` are same with the result of `df -P`.

astj:
https://github.com/mackerelio/mackerel-agent/pull/263#issuecomment-239995262

Okay, I understand 🙆

PR #268, opened 2016-08-23
remove `func (vs *Values) Merge(other Values)`

Songmu:
manipulating a pointer of map is dangerous, especially at the moment of reallocation.


itchyny:
LGTM!

PR #271, opened 2016-08-28
[incompatible] consider df  (used + available) as size of filesystem

Songmu:
The ext2/3/4 filesystems have reserved blocks which may only be allocated by privileged processes. 
Normally, the default percentage of reserved blocks is 5%.

The `1k-blocks` output by `df` command contains this reserved blocks.
So, `Used + Available` is more suited for the total available size.

http://unix.stackexchange.com/questions/41125/ext2-3-4-reserved-blocks-percentage-purpose


itchyny:
LGTM

PR #272, opened 2016-08-30
Remove DigitalOcean related comment/definition from spec/cloud.go

astj:
- Comments of `spec/cloud.go` says only EC2 is supported, but collecting metadata for GCE is already implemented
- `digitalOceanBaseURL` is defined but is not used, so removed it


Songmu:
LGTM

astj:
Thanks!

PR #274, opened 2016-09-02
Add linux distribution information to kernel spec

ak1t0:
like this

```
{
  "name":     "Linux",
  "release":  "4.4.0-36-generic",
  "version":  "#55-Ubuntu SMP Thu Aug 11 18:01:55 UTC 2016",
  "machine":  "x86_64",
  "os":       "GNU/Linux",
  "platform": "Debian 8.5",
}
```

checked distributions are as follows:
- Ubuntu
- Debian
- CentOS
- AmazonLinux
- openSUSE
- Fedora
- AlpineLinux
- ArchLinux
- Gentoo
- Oracle Linux
- Red Hat Enterprise Linux
- Scientific Linux

### DIFF
--- a/spec/linux/kernel.go
+++ b/spec/linux/kernel.go
@@ -109,7 +109,7 @@ func normalizePlatform(platform string) string {
 	case "coreos":
 		normalized = "CoreOS"
 	default:
-		normalized = ""
+		normalized = platform
 	}
 
 	return normalized
--- a/spec/linux/kernel.go
+++ b/spec/linux/kernel.go
@@ -76,6 +76,8 @@ func normalizePlatform(platform string) string {
 		normalized = "Fedora"
 	case "oracle":
 		normalized = "Oracle Linux"
+	case "enterpriseenterprise":
+		normalized = "Oracle Enterprise Linux"
 	case "centos":
 		normalized = "CentOS"
 	case "redhat":
--- a/spec/linux/kernel.go
+++ b/spec/linux/kernel.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	"github.com/mackerelio/mackerel-agent/logging"
+	"github.com/shirou/gopsutil/host"
 )
 
 // KernelGenerator XXX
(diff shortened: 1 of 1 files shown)


spec/linux/kernel.go
```diff
@@ -42,5 +43,72 @@ func (g *KernelGenerator) Generate() (interface{}, error) {
 		results[key] = str
 	}
 
+	hostInfo, err := host.Info()
+	if err != nil {
+		kernelLogger.Errorf("Failed to get platform information: %s", err)
```
haya14busa: you should return when getting err. `hostInfo` can be nil.
 
`return nil, err`

spec/linux/kernel.go
```diff
@@ -42,5 +43,72 @@ func (g *KernelGenerator) Generate() (interface{}, error) {
...
+		normalized = "Arch Linux"
+	case "exherbo":
+		normalized = "Exherbo"
+	case "alpine":
+		normalized = "Alpine Linux"
+	case "coreos":
+		normalized = "CoreOS"
+	default:
+		normalized = ""
+	}
```
haya14busa: from a quick read,  platform also can be "enterpriseenterprise", but it's not listed.
https://github.com/shirou/gopsutil/blob/859c81da50839c6829da58c9e90e1fff1edd8194/host/host_linux.go#L339

is  it ok?

spec/linux/kernel.go
```diff
@@ -42,5 +43,72 @@ func (g *KernelGenerator) Generate() (interface{}, error) {
...
+	case "arch":
+		normalized = "Arch Linux"
+	case "exherbo":
+		normalized = "Exherbo"
+	case "alpine":
+		normalized = "Alpine Linux"
+	case "coreos":
+		normalized = "CoreOS"
+	default:
+		normalized = ""
```
haya14busa: [comment]

```
		platform = strings.ToLower(lsb.ID)
		version = lsb.Release
```
https://github.com/shirou/gopsutil/blob/859c81da50839c6829da58c9e90e1fff1edd8194/host/host_linux.go#L330-L331

platform can be non-empty `lsb.ID` string, so it might be useful not to discard information.

but, i don't know what Isb.ID is like and i assume normalized name should be readable string, so it could be ok as it it.

haya14busa:
[nit] I think it's preferable to write easy-simple test (TestKernelGenerate).
You can check platform key must not contains trailing space or something.

spec/linux/kernel.go
```diff
@@ -42,5 +43,72 @@ func (g *KernelGenerator) Generate() (interface{}, error) {
 		results[key] = str
 	}
 
+	hostInfo, err := host.Info()
+	if err != nil {
+		kernelLogger.Errorf("Failed to get platform information: %s", err)
+	}
+
+	if np := normalizePlatform(hostInfo.Platform); np != "" {
+		platformInfo := np + " " + hostInfo.PlatformVersion
```
haya14busa: [comment]
Once you concatenate platform and version, you cannot get platform and version information separately (np can contains empty string), so it might be useful to  store data separately, and concatenate it when using it in view page or somewhere else.

How do you think?

spec/linux/kernel.go
ak1t0: I added `return results nil` for compatibility,  because results is not returned when `host.Info` returns error.

haya14busa: Nice! :+1:

PR #275, opened 2016-09-02
http_proxy configuration

Songmu:
- we can write http_proxy setting in mackerel-agent.conf.
- needless to write `export` in `/etc/{default,sysconfig}/mackerel-agent` for `HTTP_PROXY` setting.


itchyny:
LGTM!

PR #281, opened 2016-09-08
[nit] Remove stringer

Songmu:
disable code generation with `stringer` for avoiding dirty hack for `golint` and use `-set_exit_status` option in `golint`


astj:
LGTM

PR #283, opened 2016-09-29
Disable http2 for now

Songmu:


astj:
LGTM

(I hope this change will be reverted in near future...)

PR #285, opened 2016-10-01
don't use HTTP_PROXY when requesting cloud instance metadata APIs

Songmu:


spec/cloud.go
```diff
@@ -56,10 +56,18 @@ func SuggestCloudGenerator() *CloudGenerator {
...
 }
 
-func isEC2() bool {
-	cl := http.Client{
+func httpCli() *http.Client {
+	return &http.Client{
 		Timeout: timeout,
+		Transport: &http.Transport{
+			// don't use HTTP_PROXY when requesting cloud instance metadata APIs
+			Proxy: nil,
```
astj: [memo]
https://golang.org/pkg/net/http/#Transport

```
// If Proxy is nil or returns a nil *URL, no proxy is used.
```

PR #286, opened 2016-10-04
Add an option to output filesystem-related metrics with key by mountpoint

astj:
I'd like to add a new option to output filesystem and disk metrics with key by mountpoint, not by device. (By default this new option is disabled, so default behavior will not change.)

For example, in following environment:
```
$ df
Filesystem     1K-blocks      Used Available Use% Mounted on
udev              245912        12    245900   1% /dev
tmpfs              50172       360     49812   1% /run
/dev/sda1       41251136   1735928  37779296   5% /
none                   4         0         4   0% /sys/fs/cgroup
none                5120         0      5120   0% /run/lock
none              250856         0    250856   0% /run/shm
none              102400         0    102400   0% /run/user
vagrant        487385240 341267704 146117536  71% /vagrant
```

Original metrics are like this:
```
          "disk.sda1.reads.delta": 0,
          "disk.sda1.writes.delta": 1,
          "disk.sda.reads.delta": 0,
          "disk.sda.writes.delta": 1,
          "filesystem.sda1.size": 40463589376,
          "filesystem.sda1.used": 1777532928,
```

With the new option, following metrics will be generated(`_` is a sanitized form of `/`):
```
          "disk._.reads.delta": 0,
          "disk._.writes.delta": 1,
          "disk.sda.reads.delta": 0,
          "disk.sda.writes.delta": 1,
          "filesystem._.size": 40463589376,
          "filesystem._.used": 1777532928,
```

To enable this option, add `use_mountpoint = true` to Filesystems section of mackerel-agent.conf<del> or specify `-use_mountpoint` directly to mackere-agent</del>.
```
[Filesystems]
use_mountpoint = true
```

### DIFF
--- a/command/command_darwin.go
+++ b/command/command_darwin.go
@@ -27,7 +27,7 @@ func metricsGenerators(conf *config.Config) []metrics.Generator {
 		&metricsDarwin.CPUUsageGenerator{},
 		&metricsDarwin.MemoryGenerator{},
 		&metricsDarwin.SwapGenerator{},
-		&metrics.FilesystemGenerator{IgnoreRegexp: conf.Filesystems.Ignore.Regexp},
+		&metrics.FilesystemGenerator{IgnoreRegexp: conf.Filesystems.Ignore.Regexp, UseMountPoint: conf.Filesystems.UseMountPoint},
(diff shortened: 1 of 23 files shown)


Songmu:
LGTM

astj:
Thanks!

PR #294, opened 2016-11-16
Suppress ec2 metadata warnings

itchyny:
The metadata public-hostname/public-ipv4 are unavailable on many environment. The warning is not that important worthing recording to the mackerel-agent.log.


astj:
LGTM

itchyny:
Thank you.

PR #295, opened 2016-11-16
Uncapitalize error messages

itchyny:
We should not capitalize error strings, otherwise go vet will fail on Travis.
https://github.com/golang/go/wiki/CodeReviewComments#error-strings


astj:
LGTM!

itchyny:
Thanks!

PR #299, opened 2016-12-16
Avoid time.Tick and use time.NewTicker instead

haya14busa:
I ran https://github.com/dominikh/go-staticcheck for mackerel-agent and noticed that it uses time.Tick which is not safe to use in production code because underlying goroutine leaks.

```
$ staticcheck ./...
/home/haya14busa/src/github.com/mackerelio/mackerel-agent/agent/agent.go:43:8: using time.Tick leaks the underlying ticker, consider using it only in endless functions, tests and the main package, and use time.NewTicker here (SA1015)
```

There are no goroutine leaks caused by time.Tick in mackerel-agent because `Agent.Watch()` is called only once. However, just to be on the safe side, I think it's better to use `time.NewTicker()` and stop goroutine with quit.

godoc:
- https://golang.org/pkg/time/#Tick
- https://golang.org/pkg/time/#NewTicker

I also add a test for Agent.Watch().

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -33,25 +33,31 @@ func (agent *Agent) CollectMetrics(collectedTime time.Time) *MetricsResult {
 }
 
 // Watch XXX
-func (agent *Agent) Watch() chan *MetricsResult {
+func (agent *Agent) Watch(quit chan struct{}) chan *MetricsResult {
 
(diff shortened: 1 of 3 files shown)


Songmu:
great job!

haya14busa:
Thanks :+1:

PR #305, opened 2017-01-04
Remove duplicated generator in Windows

astj:
ref: #301 

In #301, `metrics.pluginGenerator` is added to `Agent.PluginGenerators`, but originally `metricsWindows.PluginGenerator` is already exists in `Agent.MetricsGenerators`, so they are duplicated.

As long as I see, `metrics.pluginGenerator` works fine in Windows, so I removed `metricsWindows.PluginGenerator` from `Agent.MetricsGenerators`.


astj:
I tested on my local Windows, and it seems working fine

daiksy:
LGTM!!!!!!!

astj:
Thanks!

PR #316, opened 2017-01-21
Fix segfault when loading a bad config file

hanazuki:
This patch fixes a segfault when loading a bad config file (e.g. missing mandatory key)


itchyny:
Thank you!

PR #317, opened 2017-01-24
[EXPERIMENTAL] systemd support for CentOS 7

astj:
# What I do now

- Add experimental systemd service
- Add a new make task to build systemd-ready RPM for C7

# What I don't do now

- <del>Build systemd-supported RPM in travis CI</del>
  - Maybe Ubuntu 12 cannot build systemd-supported RPM, because of lacking systemd related rpm macros
  - <del>Somehow we have to do it, but not now</del>
  - building RPMs on Docker solved this!
- Staging and kcps support
  - Possible, but later.

# Differences between this systemd RPM and current init.d RPM

- systemd cannot change path PIDFile dynamically
- systemd outputs logs to systemd default logs, not to `/var/log/mackerel-agent.log`
  - For systemd maybe this is better, because we can view logs with `sudo systemctl status mackerel-agent` ?
- <del>(Maybe) `HTTP_PROXY` is not supported yet in systemd</del>
  - Thanks to EnvironmentFile
- <del>reload not supported yet in systemd</del>
  - <del>Maybe problem with `AUTO_RETIREMENT=1`</del>
  - Thanks to supervisor mode

# References

https://fedoraproject.org/wiki/Packaging:Scriptlets?rd=Packaging:ScriptletSnippets#Systemd
https://www.freedesktop.org/software/systemd/man/systemd.unit.html (and other systemd.* )

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,8 +21,6 @@ script:
 - test `gofmt -l . | wc -l` = 0
 - make crossbuild
 after_script:
-- docker pull astj/mackerel-rpm-builder:c7
-- make rpm-systemd
 - goveralls -coverprofile=.profile.cov
 before_deploy:
 - docker pull astj/mackerel-rpm-builder:c5
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,7 @@ after_script:
 - goveralls -coverprofile=.profile.cov
 before_deploy:
 - docker pull astj/mackerel-rpm-builder:c6
+- docker pull astj/mackerel-rpm-builder:c7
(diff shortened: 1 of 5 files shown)


mattn:
Just nit sorry. mackerel-agent should start after network-online at least.
```
[Unit]
After=network-online.target

[Service]
blah-blah

[Install]
WantedBy=network-online.target
```

Songmu:
LGTM

astj:
Thanks! I reverted https://github.com/mackerelio/mackerel-agent/pull/317/commits/60f8644d82a9cb349ba9959596a17ee308c74c87 , not to run `make rpm-systemd` on each branch's tests.

PR #322, opened 2017-01-30
Refactor plugin configurations

itchyny:
The configurations of metric plugins and check plugins are contained in Config.Plugin. It makes it easy to decode from the configuration file. But basically we should differentiate the struct of metric plugin and check plugin in my opinion. A check plugin config has `NotificationInterval` field but a metric plugin config does not.
This pull request introduces `MetricPlugin` and `CheckPlugin` and now a `Config` has `MetricPlugins` and `CheckPlugins`. The method `*Config.buildPlugins` builds these plugin configurations and clears the Plugin field to stop others use this field later. I think we should remove this field if possible and move the building phase on toml unmarshalization but I'm not sure how to do.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -98,7 +98,7 @@ func (pconf *PluginConfig) buildMetricPlugin() (*MetricPlugin, error) {
 	}, nil
 }
 
-// Run the plugin
+// Run the metric plugin.
 func (pconf *MetricPlugin) Run() (string, string, int, error) {
 	if len(pconf.CommandArgs) > 0 {
 		return util.RunCommandArgs(pconf.CommandArgs, pconf.User)
@@ -140,7 +140,7 @@ func (pconf *PluginConfig) buildCheckPlugin() (*CheckPlugin, error) {
 	}, nil
 }
 
-// Run the plugin
+// Run the check plugin.
 func (pconf *CheckPlugin) Run() (string, string, int, error) {
 	if len(pconf.CommandArgs) > 0 {
 		return util.RunCommandArgs(pconf.CommandArgs, pconf.User)
--- a/config/config.go
+++ b/config/config.go
@@ -274,10 +274,38 @@ func LoadConfig(conffile string) (*Config, error) {
 	return config, err
 }
 
+func (conf *Config) makePlugins() error {
+	conf.MetricPlugins = make(map[string]*MetricPlugin)
+	if pconfs, ok := conf.Plugin["metrics"]; ok {
+		var err error
+		for name, pconf := range pconfs {
+			conf.MetricPlugins[name], err = makeMetricPlugin(pconf)
+			if err != nil {
+				return err
+			}
+		}
+	}
+	conf.CheckPlugins = make(map[string]*CheckPlugin)
+	if pconfs, ok := conf.Plugin["checks"]; ok {
+		var err error
+		for name, pconf := range pconfs {
+			conf.CheckPlugins[name], err = makeCheckPlugin(pconf)
+			if err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
 func loadConfigFile(file string) (*Config, error) {
 	config := &Config{}
 	if _, err := toml.DecodeFile(file, config); err != nil {
 		return config, err
+	}
+
+	if err := config.makePlugins(); err != nil {
+		return nil, err
 	}
 
 	if config.Include != "" {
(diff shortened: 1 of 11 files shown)


config/config_test.go
```diff
@@ -190,21 +191,31 @@ func TestLoadConfigFile(t *testing.T) {
...
-	pluginConf := config.Plugin["metrics"]["mysql"]
+	pluginConf := config.MetricPlugins["mysql"]
 	if pluginConf.Command != "ruby /path/to/your/plugin/mysql.rb" {
 		t.Errorf("plugin conf command should be 'ruby /path/to/your/plugin/mysql.rb' but %v", pluginConf.Command)
 	}
 	if pluginConf.User != "mysql" {
 		t.Errorf("plugin user_name should be 'mysql'")
 	}
+	if *pluginConf.CustomIdentifier != "app1.example.com" {
+		t.Errorf("plugin custom_identifier should be 'app1.example.com'")
```
haya14busa: it should be `t.Error` or show `*pluginConf.CustomIdentifier` as well.
Including following error messages.

config/config.go
```diff
@@ -260,16 +358,10 @@ func includeConfigFile(config *Config, include string) error {
...
 		}
 
-		for kind, plugins := range config.Plugin {
-			for key, conf := range plugins {
-				if pluginSaved[kind] == nil {
-					pluginSaved[kind] = PluginConfigs{}
-				}
-				pluginSaved[kind][key] = conf
-			}
+		// Overwrite plugin configurations.
```
itchyny: It overwrites the configuration of the same plugin name...

haya14busa: Yes, I understand that. but old code explicitly saves plugins and restores them.
So, I'm confused at first. maybe just adding "same plugin name of configurations" is fine.

config/config.go
```diff
@@ -208,25 +288,49 @@ func LoadConfig(conffile string) (*Config, error) {
...
+		for name, pconf := range pconfs {
+			conf.CheckPlugins[name], err = pconf.buildCheckPlugin()
+			if err != nil {
+				return err
+			}
+		}
+	}
+	// Make Plugins empty because we should not use this later.
+	// Use MetricPlugins and CheckPlugins.
+	conf.Plugin = nil
```
itchyny: 🆗

config/config_test.go
```diff
@@ -190,21 +191,31 @@ func TestLoadConfigFile(t *testing.T) {
 		t.Error("PostMetricsRetryMax should be 5")
 	}
 
-	if config.Plugin["metrics"] == nil {
```
itchyny: I'll add a simple test.

config/config_test.go
itchyny: For sure.

astj:
> I think we should remove this field if possible and move the building phase on toml unmarshalization but I'm not sure how to do.

Maybe we can do it by implementing `UnmarshalTOML` for `config.Plugin`?
https://godoc.org/github.com/BurntSushi/toml#Unmarshaler
https://godoc.org/github.com/BurntSushi/toml#ex-package--UnmarshalTOML

config/config.go
```diff
@@ -208,25 +289,49 @@ func LoadConfig(conffile string) (*Config, error) {
 	return config, err
 }
 
+func (conf *Config) buildPlugins() error {
```
astj: I think this name is ambiguous or misleading.
From this name, people may misunderstand it to build plugin-related configs only from `config.Plugin`. But actually this method doesn't erase already existing `conf.MetricPlugins` and `conf.CheckPlugins`, and should not erase them because it's called on `includeConfigFile`.

config/config.go
```diff
@@ -152,16 +213,36 @@ func (r *Regexpwrapper) UnmarshalText(text []byte) error {
...
 	checks := []string{}
-	for name := range conf.Plugin["checks"] {
+	for name := range conf.CheckPlugins {
 		checks = append(checks, name)
 	}
 	return checks
 }
 
-// LoadConfig XXX
+// CustomIdentifiers returns a list of customIdentifiers.
```
astj: This comment, or method name should be more verbose that this method aggregates information inside `config.MetricPlugins`. We cannot read it from current method name`CustomIdentifiers`.

itchyny:
> Maybe we can do it by implementing UnmarshalTOML for config.Plugin?

Yes, probably we have to implement unmarshalizer by hand using reflect. I think it does too much...
```go

func createTagMap(rv reflect.Type) map[string]int {
	num := rv.NumField()
	ret := make(map[string]int)
	for i := 0; i < num; i += 1 {
		field := rv.Field(i)
		name := field.Tag.Get("toml")
		if name == "" {
			name = strings.ToLower(field.Name)
		}
		ret[name] = i
	}
	return ret
}

// UnmarshalTOML defines how to decode toml data to Config
func (conf *Config) UnmarshalTOML(data interface{}) error {
	dataMap, ok := data.(map[string]interface{})
	if !ok != nil {
		return errors.New("type mismatch for Config: expected a map but found %T", data)
	}
	rv := reflect.ValueOf(v)
	if rv.IsNil() {
		return errors.New("Decode of nil %s", reflect.TypeOf(v))
	}
	tm := createTagMap(rv.Type())
	for key, datum := range dataMap {
		if key == "plugin" {
			continue
		}
		if i, ok := tm[key]; ok {
			switch rv.Field(i).Kind() {
			case reflect.String:
				// Apibase, Apikey, etc
				rv.Field(i).Set(reflect.ValueOf...
			case reflect.Struct:
				// ConnectionConfig, HostStatus, FileSystems
				Unmarshalize again and again...
		}
	}
}
```

config/config.go
itchyny: OK, I renamed to ListCustomIdentifiers.

config/config.go
itchyny: Okay, I renamed to `setMetricPluginsAndCheckPlugins` to avoid ambiguity.

astj:
LGTM!

itchyny:
Thank you!

PR #327, opened 2017-02-06
add `supervise` subcommand (supervisor mode)

Songmu:
## Usage

```shell
mackerel-agent supervise -conf ...
```

## Description

The supervisor mode launches the mackerel-agent as a child process and controls it.
In paticular, it does the following things.

- reloading mackerel-agent with HUP signal
- crash recovery
  - However, if the child process dies within 30 seconds, it is regarded as launching failure and terminate the process without crash recovery


fork.go
```diff
@@ -18,7 +20,17 @@ func handleFork(prog string, argv []string) error {
 	_ = startAt
 	cmd.Start()
 
+	c := make(chan os.Signal)
+	signal.Notify(c, os.Interrupt, syscall.SIGTERM, syscall.SIGQUIT, syscall.SIGHUP)
+	go func() {
+		for sig := range c {
+			if sig == syscall.SIGHUP {
```
mattn: syscall.SIGHUP, SIGTERM, SIGQUIT, SIGHUP doesn't work on Windows.

fork.go
```diff
@@ -23,13 +31,32 @@ func (cm *cmdManager) buildCmd() *exec.Cmd {
...
-	return cm.cmd.Wait()
+func (cm *cmdManager) stop(sig os.Signal) error {
+	cm.signaled = true
+	return cm.cmd.Process.Signal(sig)
+}
+
+func (cm *cmdManager) reload() error {
+	// TODO configtest
+	cm.hupped = true
+	return cm.cmd.Process.Signal(syscall.SIGTERM)
```
mattn: ditto

Songmu:
@mattn I have no plan to support supervise mode on windows.
Because wix/wrapper should do supervise on windows.

mattn:
Okay, then please put build-constraint to avoid compilation errors.

mattn:
Ah, sorry. I thought Windows doesn't have definition of SIGQUIT or some signals (which doesn't work on Windows). Then, your change is best for fixing my issue. Thanks!

supervisor.go
```diff
@@ -0,0 +1,106 @@
...
+)
+
+type supervisor struct {
+	prog string
+	argv []string
+
+	cmd      *exec.Cmd
+	startAt  time.Time
+	signaled bool
+	hupped   bool
```
haya14busa: Isn't it better to guard signaled and hupped with mutex?

supervisor.go
```diff
@@ -0,0 +1,106 @@
...
+				logger.Warningf("failed to reload: %s", err.Error())
+			}
+		} else {
+			sv.stop(sig)
+		}
+	}
+}
+
+func (sv *supervisor) supervise() error {
+	sv.start()
```
haya14busa: maybe it's better to check err from `sv.start()` here in case the command is something wrong.

supervisor_test.go
```diff
@@ -0,0 +1,181 @@
...
+	pid := sv.cmd.Process.Pid
+	if !existsPid(pid) {
+		t.Errorf("process doesn't exist")
+	}
+	time.Sleep(50 * time.Millisecond)
+	ch <- os.Interrupt
+
+	err := <-done
+	if err != nil {
+		t.Errorf("error should be nil but: %s", err)
```
haya14busa: [nit] should be `%v` for err format?

supervisor.go
haya14busa: and we should run go test with -race in CI, i guess...

supervisor.go
haya14busa: and we should run go test with -race in CI, i guess...

supervisor_test.go
```diff
@@ -0,0 +1,181 @@
...
+		done <- sv.wait()
+	}()
+	pid := sv.cmd.Process.Pid
+	if !existsPid(pid) {
+		t.Errorf("process doesn't exist")
+	}
+	time.Sleep(50 * time.Millisecond)
+	ch <- os.Interrupt
+
+	err := <-done
```
haya14busa: we can remove `done` and just use `<- sv.wait()` instead.

supervisor.go
Songmu: fixed at 1356a53

supervisor.go
Songmu: fixed at a225f87

supervisor_test.go
Songmu: fixed at 52d2cb1

supervisor_test.go
Songmu: I changed interface of `supervisor.supervise()` at a85cbd2.

How about it?

supervisor/supervisor.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+// Supervise the mackerel-agent
+func (sv *Supervisor) Supervise(c chan os.Signal) error {
+	err := sv.start()
+	if err != nil {
+		return err
+	}
+	if c == nil {
+		c = make(chan os.Signal, 1)
```
haya14busa: it's better to close `c` not to leak `go sv.handleSignal(c)` goroutine.

```go
defer close(c)
```

I also confirmed there is a leak with https://github.com/fortytw2/leaktest

Also, it's better to  document `c chan os.Signal` arg can be nil and if it's not nil, the caller has responsibility to close it. Or, we can just say "do not use this package as library", i guess?

e.g. 

```go
// Supervise starts a child mackerel-agent process and supervises it.
// 'c' can be nil and it's tipically nil. When you pass signal channel to this
// method, you have to close the channel to stop internal goroutine.
func (sv *Supervisor) Supervise(c chan os.Signal) error {
```

supervisor/supervisor.go
```diff
@@ -0,0 +1,164 @@
...
+				logger.Warningf("failed to reload: %s", err.Error())
+			}
+		} else {
+			sv.stop(sig)
+		}
+	}
+}
+
+// Supervise the mackerel-agent
+func (sv *Supervisor) Supervise(c chan os.Signal) error {
```
haya14busa: Interface can be something like this and it looks simple...?

```go
func Supervise(agentProg string, argv []string, c chan os.Signal) error {
```

Also, with this interface, we don't have to care the caller side modifies sv.Prog or sv.ArgV.

But, maybe.... it's ok as it is since we uses `*Supervisor` in test and we don't have to care caller side so much because we expects it's not used as library (right?)....

util/pid.go
```diff
@@ -0,0 +1,6 @@
+package util
```
Songmu: fixed at bc8bba7.

util/pid.go
```diff
@@ -0,0 +1,6 @@
...
+
+// ExistsPid check if pid exists
+func ExistsPid(pid int) bool {
```
Songmu: fixed at e13d0fd

supervisor/supervisor.go
```diff
@@ -0,0 +1,164 @@
...
+	err := sv.start()
```
Songmu: fixed at e13d0fd

supervisor/supervisor.go
Songmu: fixed at 5966779.

supervisor/supervisor.go
Songmu: fixed at 5966779.

pidfile/pidfile_test.go
```diff
@@ -0,0 +1,61 @@
+// +build !windows
```
astj: For which reason this test is skipped on windows?
`Create` and `Remove` is performed in windows too, so it's better test them on windows

astj: I guess this skip is due to missing implementation of `existsPid`, but even so it's better there's some tests for Windows (maybe in another file like `pidfile_windows_test.go`).

supervisor/supervisor.go
```diff
@@ -0,0 +1,171 @@
...
+func (sv *supervisor) handleSignal(ch <-chan os.Signal) {
+	for sig := range ch {
+		if sig == syscall.SIGHUP {
+			logger.Infof("receiving HUP, spawning a new mackerel-agent")
+			err := sv.reload()
+			if err != nil {
+				logger.Warningf("failed to reload: %s", err.Error())
+			}
+		} else {
+			sv.stop(sig)
```
astj: [imo]
Isn't it better `supervisor` outputs a log here like L149?

supervisor/supervisor.go
```diff
@@ -0,0 +1,171 @@
...
+		}
+	}
+}
+
+func (sv *supervisor) supervise(c chan os.Signal) error {
+	if err := sv.start(); err != nil {
```
astj: [imo]
same as L155

supervisor/supervisor.go
astj: (Especially here, logging which signal the supervisor received is maybe good)

supervisor/supervisor.go
```diff
@@ -0,0 +1,171 @@
+package supervisor
```
astj: Since supervisor.go is not used in Windows, so is it possible to add `// +build !windows` and add dummy supervisor_windows.go?
(I don't know the best way. One might think current way is better because of simplicity).

supervisor/testdata/stub-agent.go
```diff
@@ -0,0 +1,45 @@
+package main
```
astj: Cute stub ♥️

PR #331, opened 2017-02-09
support metadata plugins in configuration

itchyny:
We plan to add metadata plugin feature in agent. This pull request adds support for read the configuration of metadata plugins from the config file.


Songmu:
LGTM

PR #332, opened 2017-02-13
Do not create pidfile unless explicitly specified

Songmu:
Currently, mackerel-agent implicitly creates pidfile even if not specified in argument or configuration file.

I think it is not good behaviour, so I fixed not to create pidfile unless explicitly specified.

It is slightly imcompatible change, but it does not affect much. Because `-pidfile` option is explicitly specified in init scripts in the rpm/deb.

see. https://github.com/mackerelio/mackerel-agent/tree/3aa7cbaec4b68b374b196defd24c8512131bbf6d/packaging


astj:
> but it does not affect much.

It does to macOS, because `launchctl` plist doesn't specify `-pidfile` option.
https://github.com/mackerelio/homebrew-mackerel-agent/blob/master/mackerel-agent.rb#L50-L54
To keep current behavior in macOS, we need to modify the plist.
Or we can stop using pidfile on macOS and leave the process management of `mackerel-agent` to `launchctl`?

config/config_bsd.go
```diff
@@ -7,11 +9,10 @@ import (
 
 var mackerelRoot = filepath.Join(os.Getenv("HOME"), "Library", getAgentName())
```
astj: Current {free,net}bsd also uses this directory as root directory, but I think this is not good for {free,net}bsd.
(I guess current `mackerelRoot` of {free,net}bsd is just copied from macOS)
Isn't it better to set somewhere more *nix-friendly one?

PR #333, opened 2017-02-15
Add metadata plugin feature

itchyny:
previous pull request: #331 

This pull request adds support for metadata plugin. User now can configure metadata plugins and mackerel-agent posts the metadata periodically.

### DIFF
--- a/command/metadata.go
+++ b/command/metadata.go
@@ -1,0 +1,21 @@
+package command
+
+import (
+	"github.com/mackerelio/mackerel-agent/config"
+	"github.com/mackerelio/mackerel-agent/metadata"
+)
+
+func metadataGenerators(conf *config.Config) []*metadata.Generator {
+	generators := make([]*metadata.Generator, 0, len(conf.MetadataPlugins))
+
+	for name, pluginConfig := range conf.MetadataPlugins {
+		generator := &metadata.Generator{
+			Name:   name,
+			Config: pluginConfig,
+		}
+		logger.Debugf("Metadata plugin generator created: %v", generator)
+		generators = append(generators, generator)
+	}
+
+	return generators
+}
--- a/command/metadata.go
+++ b/command/metadata.go
@@ -51,12 +51,13 @@ func runMetadataLoop(c *Context, termMetadataCh <-chan struct{}, quit <-chan str
 		}
 
 		results := []*metadataResult{}
+	ConsumeResults:
 		for {
 			select {
 			case result := <-resultCh:
 				results = append(results, result)
 			default:
-				break
+				break ConsumeResults
 			}
 		}
 
--- a/command/metadata.go
+++ b/command/metadata.go
@@ -51,13 +51,12 @@ func runMetadataLoop(c *Context, termMetadataCh <-chan struct{}, quit <-chan str
 		}
 
 		results := []*metadataResult{}
-		hasResult := true
-		for hasResult {
+		for {
 			select {
 			case result := <-resultCh:
 				results = append(results, result)
 			default:
-				hasResult = false
+				break
 			}
 		}
 
--- a/command/metadata.go
+++ b/command/metadata.go
@@ -1,6 +1,8 @@
 package command
 
 import (
+	"time"
+
 	"github.com/mackerelio/mackerel-agent/config"
 	"github.com/mackerelio/mackerel-agent/metadata"
 )
@@ -18,4 +20,76 @@ func metadataGenerators(conf *config.Config) []*metadata.Generator {
 	}
 
 	return generators
+}
+
+type metadataResult struct {
+	namespace string
+	metadata  interface{}
+}
+
+func runMetadataLoop(c *Context, termMetadataCh <-chan struct{}, quit <-chan struct{}) {
+	resultCh := make(chan *metadataResult)
+	for _, g := range c.Agent.MetadataGenerators {
+		go runEachMetadataLoop(g, resultCh, quit)
+	}
+
+	exit := false
+	for !exit {
+		select {
+		case <-time.After(1 * time.Minute):
+		case <-termMetadataCh:
+			logger.Debugf("received 'term' chan for metadata loop")
+			exit = true
+		}
+
+		results := []*metadataResult{}
+		hasResult := true
+		for hasResult {
+			select {
+			case result := <-resultCh:
+				results = append(results, result)
+			default:
+				hasResult = false
+			}
+		}
+
+		for _, result := range results {
+			err := c.API.PutMetadata(c.Host.ID, result.namespace, result.metadata)
+			if err != nil {
+				logger.Errorf("put metadata %q failed: %s", result.namespace, err.Error())
+				// TODO: error handling (re-enqueue?)
+			}
+		}
+	}
+}
+
+func runEachMetadataLoop(g *metadata.Generator, resultCh chan<- *metadataResult, quit <-chan struct{}) {
+	interval := g.Interval()
+	nextInterval := 10 * time.Second
+	nextTime := time.Now()
+
+	for {
+		select {
+		case <-time.After(nextInterval):
+			metadata, err := g.Fetch()
+
+			// case for laptop sleep mode (now >> nextTime + interval)
+			now := time.Now()
+			nextInterval = interval - (now.Sub(nextTime) % interval)
+			nextTime = now.Add(nextInterval)
+
+			if err != nil {
+				logger.Warningf("metadata %q: %s", g.Name, err.Error())
+				continue
+			}
+			logger.Debugf("metadata %q: %v", g.Name, metadata)
+			resultCh <- &metadataResult{
+				namespace: g.Name,
+				metadata:  metadata,
+			}
+
+		case <-quit:
+			return
+		}
+	}
 }
--- a/command/metadata.go
+++ b/command/metadata.go
@@ -106,7 +106,7 @@ func runEachMetadataLoop(g *metadata.Generator, resultCh chan<- *metadataResult,
 				continue
 			}
 
-			if err = g.Save(metadata); err != nil {
+			if err := g.Save(metadata); err != nil {
 				logger.Warningf("metadata plugin %q: %s", g.Name, err.Error())
 				continue
 			}
--- a/command/metadata.go
+++ b/command/metadata.go
@@ -101,7 +101,7 @@ func runEachMetadataLoop(g *metadata.Generator, resultCh chan<- *metadataResult,
 				continue
 			}
 
-			if !g.Differs(metadata) {
+			if !g.IsChanged(metadata) {
 				logger.Debugf("skipping metadata %q: %v", g.Name, metadata)
 				continue
 			}
--- a/command/metadata.go
+++ b/command/metadata.go
@@ -19,9 +19,9 @@ func metadataGenerators(conf *config.Config) []*metadata.Generator {
 
 	for name, pluginConfig := range conf.MetadataPlugins {
 		generator := &metadata.Generator{
-			Name:     name,
-			Config:   pluginConfig,
-			Tempfile: filepath.Join(workdir, "mackerel-metadata", name),
+			Name:      name,
+			Config:    pluginConfig,
+			Cachefile: filepath.Join(workdir, "mackerel-metadata", name),
 		}
 		logger.Debugf("Metadata plugin generator created: %#v %#v", generator, generator.Config)
 		generators = append(generators, generator)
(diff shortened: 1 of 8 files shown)


metadata/metadata.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+	g.PrevMetadata = metadata
+}
+
+// Save stores the metadata locally
+func (g *Generator) Save(metadata interface{}) error {
+	g.PrevMetadata = metadata
+	data, err := json.Marshal(metadata)
+	if err != nil {
+		return fmt.Errorf("failed to marshal the metadata to json: %v %s", metadata, err.Error())
```
haya14busa: metadata may be  large, so I think it's better to place error message before metadata.

metadata/metadata.go
```diff
@@ -0,0 +1,123 @@
...
+	tmpf, err := ioutil.TempFile("", "")
+	if err != nil {
+		return err
+	}
+	defer os.Remove(tmpf.Name())
+	_, err = tmpf.Write(contents)
+	if err != nil {
+		return err
+	}
+	tmpf.Close()
```
haya14busa: [nit] if "Write" fails, tmpf won't be closed, so I think you can call `defer tmpf.Close()` before `defer os.Remove(tmpf.Name())` line

metadata/metadata_test.go
```diff
@@ -0,0 +1,181 @@
...
+		var metadata interface{}
+		_ = json.Unmarshal([]byte(test.metadata), &metadata)
+
+		got := g.Differs(metadata)
+		if got != test.differs {
+			t.Errorf("Differs() should return %t but got %t for %v, %v", test.differs, got, prevmetadata, metadata)
+		}
+
+		if err := g.Clear(); err != nil {
+			t.Errorf("Error should not occur in Clear() but got: %s", err.Error())
```
haya14busa: [nit] it's common to use '%v' to format error, so you don't have to call `err.Error()` and use '%s' everywhere.

metadata/metadata.go
mattn: ditto https://github.com/mackerelio/mackerel-agent/pull/333/files/e7158937fd76b893f23dd9c0560ffb0edab9abef#diff-b3a9f1c1b5d5600f4924efe942b13fb0R45

mattn: replace left & right https://github.com/mackerelio/mackerel-agent/pull/333/files/e7158937fd76b893f23dd9c0560ffb0edab9abef#diff-b3a9f1c1b5d5600f4924efe942b13fb0R85

metadata/metadata_test.go
mattn: tips:
```go
package main

import (
	"fmt"
	"io"
)

func main() {
	var err error
	
	err = io.EOF
	fmt.Printf("%v\n", err) // EOF
	err = nil
	fmt.Printf("%v\n", err) // <nil>

	err = io.EOF
	fmt.Printf("%s\n", err) // EOF
	err = nil
	fmt.Printf("%s\n", err) // %s!s(<nil>)
}
```

command/metadata.go
```diff
@@ -0,0 +1,125 @@
...
+			default:
+				hasResult = false
+			}
+		}
+
+		for _, result := range results {
+			err := c.API.PutMetadata(c.Host.ID, result.namespace, result.metadata)
+			if err != nil {
+				logger.Errorf("put metadata %q failed: %s", result.namespace, err.Error())
+				// do not retry because huge metadata will be always rejected
```
astj: Maybe we should retry unless the failure is caused by metadata size (maybe in case the response code is 400?)

command/metadata.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/mackerelio/mackerel-agent/config"
+	"github.com/mackerelio/mackerel-agent/metadata"
+)
+
+func metadataGenerators(conf *config.Config) []*metadata.Generator {
+	generators := make([]*metadata.Generator, 0, len(conf.MetadataPlugins))
+
+	workdir := os.Getenv("MACKEREL_PLUGIN_WORKDIR")
+	if workdir == "" {
+		workdir = os.TempDir()
```
astj: If many metadata plugins are prepared without `MACKEREL_PLUGIN_WORKDIR`, many tempfiles will be created to `os.TempDir`. Is it better to create a subdir and place the tempfiles into it?

command/metadata.go
```diff
@@ -0,0 +1,125 @@
...
+		for hasResult {
+			select {
+			case result := <-resultCh:
+				results = append(results, result)
+			default:
+				hasResult = false
+			}
+		}
+
+		for _, result := range results {
```
astj: Current implementation waits until all metadata plugins' execution before PUTing api, but is it necessary?
(IMO we can PUT one request immediate after receiving one result)

command/metadata.go
```diff
@@ -0,0 +1,125 @@
...
+				clearMetadataCache(c.Agent.MetadataGenerators, result.namespace)
+				continue
+			}
+		}
+	}
+}
+
+func clearMetadataCache(generators []*metadata.Generator, namespace string) {
+	for _, g := range generators {
+		if g.Name == namespace {
```
astj: I know it's necessary, but not cool…… 🤔

command/metadata.go
```diff
@@ -0,0 +1,125 @@
...
+		select {
+		case <-time.After(1 * time.Minute):
+		case <-termMetadataCh:
+			logger.Debugf("received 'term' chan for metadata loop")
+			exit = true
+		}
+
+		results := []*metadataResult{}
+		hasResult := true
+		for hasResult {
```
itchyny: Right. done: https://github.com/mackerelio/mackerel-agent/pull/333/commits/11faff8000310442412cd30dc5c0fe7c52a439a9

command/metadata.go
```diff
@@ -0,0 +1,125 @@
...
+				logger.Warningf("metadata plugin %q: %s", g.Name, err.Error())
+				continue
+			}
+
+			if !g.Differs(metadata) {
+				logger.Debugf("skipping metadata %q: %v", g.Name, metadata)
+				continue
+			}
+
+			if err = g.Save(metadata); err != nil {
```
itchyny: For sure. done: https://github.com/mackerelio/mackerel-agent/pull/333/commits/b8f33df7710fe35b0a41fcf4ee024b481d989a9f

metadata/metadata.go
```diff
@@ -0,0 +1,123 @@
...
+	var metadata interface{}
+	if err := json.Unmarshal([]byte(message), &metadata); err != nil {
+		return nil, fmt.Errorf("outputs invalid JSON: %v", message)
+	}
+
+	return metadata, nil
+}
+
+// Differs returns whether the metadata has been changed or not
+func (g *Generator) Differs(metadata interface{}) bool {
```
itchyny: Hmm, in precise, the (state of) metadata is changed (new metadata differs from previous metadata), and not the generator is changed. Looks oddly a bit for me but I changed to IsChanged https://github.com/mackerelio/mackerel-agent/pull/333/commits/090911319450e92873d25fc639c74dc1c33375b3

metadata/metadata.go
itchyny: > metadata may be large

Right. I thought it's annoying the huge (100kb in maximum) metadata to the log message so removed. https://github.com/mackerelio/mackerel-agent/commit/ef1d6a46a8199ce3cbc53ddf2da176258e0654e7

metadata/metadata.go
itchyny: done: https://github.com/mackerelio/mackerel-agent/commit/742721c1373660a14d31ba7cde56069825b0c859

mattn:
FYI, if you hate labeled-break...

```go
	for {
		select {
		case result := <-resultCh:
			results = append(results, result)
			continue
		default:
		}
		break
	}
```

command/metadata.go
itchyny: I'm not sure sub-directory is better. Can it cause another problem due to invalid permission for the directory? Well, check plugin uses `os.TempDir` for reference. https://github.com/mackerelio/go-check-plugins/blob/6481e313ae25776b1dc052490dbc4bddcb50cf73/check-log/check-log.go#L129-L132

mattn:
LGTM for fixes about I mentioned. :+1:

command/metadata.go
itchyny: Hmm, we may retry on network errors but should not on errors which will occur again and again; 413 (huge metadata), or 400 (too much metadata per host), 403 (unauthorized) ...

command/metadata.go
astj: `os.TempDir()` is shared with other processes, so putting many files directly under TempDir looks not good, just imo.

itchyny: Done: https://github.com/mackerelio/mackerel-agent/commit/7491fa0d0706b442fe38535cd997e0f461d9bde5

itchyny: Plus: https://github.com/mackerelio/mackerel-agent/commit/40e1882b2a06f8d7f480bf597031b58b877403c0

command/metadata.go
itchyny: Retry on 5XX errors: https://github.com/mackerelio/mackerel-agent/commit/17a51cbc51becbbcedbbd28e035254349afda83c and avoid infinite retry: https://github.com/mackerelio/mackerel-agent/pull/333/commits/4c6a3d6fbf38ca550d77654f22d0496bb5e56228

metadata/metadata.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/mackerelio/mackerel-agent/logging"
+)
+
+var logger = logging.GetLogger("metadata")
+
+// Generator generates metadata
+type Generator struct {
+	Name         string
+	Config       *config.MetadataPlugin
+	Tempfile     string
```
Songmu: [nit] I think `Tempfile` is not a best naming. Although it may be created in tmpdir, but It is not "temporary" file, and exists for a long time. If you have good idea, rename this field name.

command/metadata.go
```diff
@@ -0,0 +1,136 @@
...
+			exit = true
+		}
+
+		results := make(map[string]*metadataResult)
+	ConsumeResults:
+		for {
+			select {
+			case result := <-resultCh:
+				// prefer new result to old metadata which remains to be posted on retry
+				// and also avoid infinite number of retries
```
Songmu: Newer metadata is not necessarily after the queue, isn't it?

command/metadata.go
```diff
@@ -0,0 +1,136 @@
...
+				break ConsumeResults
+			}
+		}
+
+		for _, result := range results {
+			resp, err := c.API.PutMetadata(c.Host.ID, result.namespace, result.metadata)
+			// retry on 5XX errors
+			if resp != nil && resp.StatusCode >= 500 {
+				logger.Errorf("put metadata %q failed: status %s", result.namespace, resp.Status)
+				resultCh <- &metadataResult{
```
Songmu: You should use goroutine here. Otherwise it causes a deadlock. Because it will send and receive to one channel within the same thread.

itchyny: Done: https://github.com/mackerelio/mackerel-agent/pull/333/commits/22d94fe90ebf59ae4814786ab939ea241ec3307b

command/metadata.go
itchyny: Fixed: https://github.com/mackerelio/mackerel-agent/pull/333/commits/f31a1ec92e9116bbbe4ac0c001f923519ca3052c

metadata/metadata.go
itchyny: Sure. I renamed to Cachefile because the file caches the metadata locally and can be purged away. https://github.com/mackerelio/mackerel-agent/pull/333/commits/68e973fcb9c079abb62ae32d0e87ea982e9df43e

Songmu: :+1:

command/metadata.go
```diff
@@ -0,0 +1,142 @@
...
+			exit = true
+		}
+
+		results := make(map[string]*metadataResult)
+	ConsumeResults:
+		for {
+			select {
+			case result := <-resultCh:
+				// prefer new result to avoid infinite number of retries
+				if prev, ok := results[result.namespace]; ok {
```
Songmu: [nit] How about `if prev, ok := results[result.namespace]; !ok || result.createdAt.After(prev.createdAt) {` ?

Songmu: You can keep it as long as it does not fit your style.

Songmu:
LGTM!

command/metadata.go
itchyny: Thanks, I'll keep the code because I feel the if statement a kind of an idiom in Golang.

itchyny:
Thank you all for your review.

mattn:
Ah, sorry, I'm not sure, but [resultCh](https://github.com/mackerelio/mackerel-agent/blob/f31a1ec92e9116bbbe4ac0c001f923519ca3052c/command/metadata.go#L40) does not need to grow-up size?

PR #334, opened 2017-02-16
Use Named Result Parameters as document

haya14busa:
https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters

`(string, string, int, error)` is unclear.


syou6162:
LGTM!

haya14busa:
Thanks!

PR #336, opened 2017-02-18
hw.physmem is obsolete

miwarin:
hw.physmem is singned 32bit. (less 2GB)
hw.physmem64 is singned 64bit.

### DIFF
--- a/spec/netbsd/memory.go
+++ b/spec/netbsd/memory.go
@@ -30,7 +30,7 @@ const bytesInKibibytes = 1024
 func (g *MemoryGenerator) Generate() (interface{}, error) {
 	spec := map[string]string{}
 
-	cmd := exec.Command("sysctl", "-n", "hw.physmem")
+	cmd := exec.Command("sysctl", "-n", "hw.physmem64")
 	outputBytes, err := cmd.Output()
 	if err != nil {
 		return nil, fmt.Errorf("sysctl -n hw.physmem: %s", err)


mattn:
I don't know netbsd as well. So I don't make sure.

```go
cmd := exec.Command("sysctl", "-n", "hw.physmem64")
outputBytes, err := cmd.Output()
if err != nil {
	cmd = exec.Command("sysctl", "-n", "hw.physmem")
	outputBytes, err = cmd.Output()
}
if err != nil {
  		return nil, fmt.Errorf("sysctl -n hw.physmem: %s", err)
}
```

Do you think that don't need fallbacks to hw.physmem?

miwarin:
hw.physmem is unnecessary.
Because physmem64 was implemented more than 10 years ago.

mattn:
I now confirmed this on NetBSD 7.0.2 using VirtualBox. It seems having both hw.physmem, hw.physmem64. So it's better to you hw.pysmem64 as your said.

spec/netbsd/memory.go
```diff
@@ -30,7 +30,7 @@ const bytesInKibibytes = 1024
 func (g *MemoryGenerator) Generate() (interface{}, error) {
 	spec := map[string]string{}
 
-	cmd := exec.Command("sysctl", "-n", "hw.physmem")
+	cmd := exec.Command("sysctl", "-n", "hw.physmem64")
 	outputBytes, err := cmd.Output()
 	if err != nil {
 		return nil, fmt.Errorf("sysctl -n hw.physmem: %s", err)
```
astj: [nit]
Please change this error message too.

astj:
@miwarin (fyi @mattn) Thank you!
It seems good, but I left a very nit comment.  Please check :pray:

astj:
@miwarin :point_up: I fixed the error message in #343 and merged. Thanks!

PR #338, opened 2017-02-19
Improve darwin cpu spec

astj:
Currently mackerel-agent in Darwin reports only 1 CPU core even if the host has many.
And, currently only `model_name` is collected.

With this patch, agent can report multiple CPU cores if the host has so, and it also reports `family`, `model`, `stepping`, `vendor_id` specs like linux does.

Demo
```json
      "cpu": [
        {
          "cores": "2",
          "family": "6",
          "model": "61",
          "model_name": "Intel(R) Core(TM) i5-5287U CPU @ 2.90GHz",
          "stepping": "4",
          "vendor_id": "GenuineIntel"
        },
        {
          "cores": "2",
          "family": "6",
          "model": "61",
          "model_name": "Intel(R) Core(TM) i5-5287U CPU @ 2.90GHz",
          "stepping": "4",
          "vendor_id": "GenuineIntel"
        }
      ],
```

### DIFF
--- a/spec/darwin/cpu.go
+++ b/spec/darwin/cpu.go
@@ -3,7 +3,10 @@
 package darwin
 
 import (
+	"bufio"
+	"bytes"
 	"os/exec"
+	"strings"
 
 	"github.com/mackerelio/mackerel-agent/logging"
 )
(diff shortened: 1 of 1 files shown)


spec/darwin/cpu.go
```diff
@@ -21,6 +25,54 @@ var cpuLogger = logging.GetLogger("spec.cpu")
...
+		}
+	}
+
+	if err := scanner.Err(); err != nil {
+		cpuLogger.Errorf("Failed (skip this spec): %s", err)
+		return nil, err
+	}
+
+	wholeResults := []cpuSpec{}
+	for i := 0; int64(i) < cores; i++ {
```
itchyny: You can use `int` for `cores` (I do prefer coreCount) then `int64(i)` can be `i`. For conversion, `strconv.Atoi` converts string to int (not int64).

spec/darwin/cpu.go
```diff
@@ -21,6 +25,54 @@ var cpuLogger = logging.GetLogger("spec.cpu")
...
+			}
+		}
+	}
+
+	if err := scanner.Err(); err != nil {
+		cpuLogger.Errorf("Failed (skip this spec): %s", err)
+		return nil, err
+	}
+
+	wholeResults := []cpuSpec{}
```
itchyny: Allocating necessary size of slice is preferable (`make([]cpuSpec{}, 0, coreCount)`) when its capacity is known before appending.

mattn: Assignment may be better than append.

```go
wholeResults := make([]cpuSpec{}, coreCount)
for i := range wholeResults {
    wholeResults[i] = results
}
```

spec/darwin/cpu.go
```diff
@@ -66,9 +66,9 @@ func (g *CPUGenerator) parseSysCtlBytes(res []byte) (interface{}, error) {
 		return nil, err
 	}
 
-	wholeResults := []cpuSpec{}
-	for i := 0; int64(i) < cores; i++ {
-		wholeResults = append(wholeResults, results)
+	wholeResults := make([]cpuSpec, 0, coreCount)
```
mattn: make([]cpuSpec, coreCount)

astj: Thanks :pray:

itchyny:
LGTM

astj:
Thanks!

PR #339, opened 2017-02-21
run test with -race in CI

haya14busa:
gotestcover is depricated and it's not merge cover profiles properly

- https://github.com/pierrre/gotestcover
- https://github.com/haya14busa/goverage

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,7 @@ crossbuild: deps
 		-os="linux darwin freebsd netbsd" -arch="386 amd64 arm" -d . -n $(MACKEREL_AGENT_NAME)
 
 cover: deps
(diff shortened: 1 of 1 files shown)


haya14busa:
At first, I tried to replace gotestcover with gocoverage because gotestcover has downside like it doesn't support -coverpkg arg, but replacing it with goverage somehow went wrong and the coverage was decreased (even though I cannot reproduce it locallly)... so I stopped trying to replace gotestcover with goverage. Just add `-race` flag. 

I changed -covermode to atomic because 'count' doesn't work with -race. ref: golang/go#12118

Songmu:
Thanks!

PR #340, opened 2017-02-21
Fix format verb: use '%v'

haya14busa:
I found bad formatted log in test log output https://travis-ci.org/mackerelio/mackerel-agent/builds/203777712

Example:

2017/02/21 05:12:30 agent.go:110: DEBUG <agent> Failed to fetch meta information from plugin &{%!s(*config.MetricPlugin=&{../example/metrics-plugins/dice-with-meta.rb []  <nil>}) %!s(*metrics.pluginMeta=<nil>)} (non critical); seems that this plugin does not have meta information: while reading the first line of command ../example/metrics-plugins/dice-with-meta.rb: EOF


Songmu:
LGTM

PR #345, opened 2017-02-24
fix comment of supervisor

haya14busa:
The comment meant following code

	if c == nil {
		c = make(chan os.Signal, 1)
		defer close(c)
	}

but it actually looks like this

	if c == nil {
		c = make(chan os.Signal, 1)
	}
	defer close(c)

I think both are fine but comment should be true.

ref: https://github.com/mackerelio/mackerel-agent/pull/327#discussion_r102206365


Songmu:
:bow: Thanks!

PR #346, opened 2017-03-01
Keep environment variables when Agent runs commands with sudo

astj:
Especially when executing plugins with `user = ` option, currently some environment variables set by agent are dropped.
- `init_unix.go` does `setenv`
- `MACKEREL_AGENT_PLUGIN_META=1` on getting graph definitions

`sudo`ing with `-E` maybe solve those problem.
```
     -E, --preserve-env
                 Indicates to the security policy that the user wishes to preserve their existing environment variables.  The security policy may return an error if the user does not have permission
                 to preserve the environment.
```


itchyny:
LGTM!

astj:
Thanks!

PR #348, opened 2017-03-06
Fix disk metrics on Windows

mattn:
Performance Counder doesn't provide counter for collecting IOPS per disks.
This change collect metrics using WMI.


daiksy:
LGTM!!!

PR #347, opened 2017-03-03
Release systemd RPMs to github releases

astj:
- Make `make rpm{,-stage,-kcps}` to build systemd RPMs too
- Update changelog in systemd RPM in release process (same as original RPM)
- Some refactor

# Discussion

Currently mackerel-agent-systemd.spec has no changelogs before 0.40.1.  Should we add them?

PR #344, opened 2017-02-22
Build RPM files on CentOS5 on Docker

astj:
ref: #330 

Sadly CentOS 5 cannot install RPMs built on CentOS 6.
Since CentOS 6/7 can install RPMs built on CentOS 5, we build RPMs on CentOS 5 docker image.

PR #330, opened 2017-02-08
Build RPM packages with Docker

astj:
related: #325 

Though we can build current rpm in travis directly, we cannot build systemd-based one (under development on #317).
So I'd like to introduce docker based rpm build procedure.

PR #349, opened 2017-03-07
Release version 0.40.1

stanaka:
- [EXPERIMENTAL] systemd support for CentOS 7 #317
- add `supervise` subcommand (supervisor mode) #327
- Build RPM packages with Docker #330
- run test with -race in CI #339
- Use hw.physmem64 instead of hw.physmem in NetBSD #343
- Build RPM files on CentOS5 on Docker #344
- Keep environment variables when Agent runs commands with sudo #346
- Release systemd RPMs to github releases #347
- Fix disk metrics on Windows #348

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,18 @@
 # Changelog
 
+## 0.40.1 (2017-03-07)
+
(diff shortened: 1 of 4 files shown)


astj:
I'll bump to 0.41.0, not 0.40.1!

PR #357, opened 2017-03-14
Don't raise error when creating pidfile if the contents of pidfile is same as own pid

Songmu:


pidfile/pidfile_test.go
```diff
@@ -31,7 +32,16 @@ func TestCreate(t *testing.T) {
...
+}
 
+func TestCreate_mutex(t *testing.T) {
+	dir, err := ioutil.TempDir("", "")
+	if err != nil {
+		t.Fatalf("failed to create tempdir")
+	}
+	defer os.RemoveAll(dir)
+	pidfile := filepath.Join(dir, "pidfile")
+	ioutil.WriteFile(pidfile, []byte(fmt.Sprintf("%d", os.Getppid())), 0644)
```
haya14busa: If I understand correctly, `s/os.Getppid()/os.Getpid()/` and `err` should be nil.

pidfile/pidfile.go
```diff
@@ -19,6 +19,9 @@ func Create(pidfile string) error {
 	}
 	if pidString, err := ioutil.ReadFile(pidfile); err == nil {
 		if pid, err := strconv.Atoi(string(pidString)); err == nil {
+			if pid == os.Getpid() {
```
haya14busa: I think it's better to leave some comments here to describe the background of this change.

PR #358, opened 2017-03-16
Exclude _tools from package

itchyny:
Currently the packages include `_tools` file. The file looks like an executable of `gen_commands.go` and is unintentionally included in the packages. This pull request fixes the problem.


astj:
LGTM!

itchyny:
Thank you!

PR #361, opened 2017-03-22
Timeout for command execution on Windows

mattn:
songmu/timeout terminate process and the children. So this change is enough to do same as UNIX part on Windows.


Songmu:
Thanks!

PR #363, opened 2017-03-24
sh -c "exec ..."

Songmu:
If we don't use `exec` when calling external command, the plugin command is called as a child process of `sh -c`, so there is a possibility that an orphan process will occur at timeout.


Songmu:
If we don't use `exec` when calling external command, the plugin command is called as a child process of `sh -c`, so there is a possibility that an orphan process will occur at timeout.

Songmu:
It doesn't work when using pipe, redirection, etc...

PR #364, opened 2017-03-24
It need to read output from command continuously.

mattn:
Fixes to read output from command continuously because plugin may hang when agent doesn't read. For example, reading too long lines using scanner.Scan() .


Songmu:
Thanks!

PR #367, opened 2017-03-27
Logs that are not via the mackerel-agent's logger are also output to the eventlog

Songmu:
behavior is changed at https://github.com/mackerelio/mackerel-agent/pull/364, so fixed.


mattn:
Thank you for looking this. LGTM.

syou6162:
LGTM!

PR #360, opened 2017-03-21
[EXPERIMENTAL] Add systemd support for deb packages

astj:
ref: #317 

This will add new deb package for systemd-based distributions.

# This p-r introduces:

- add tenporary make task `make deb-systemd` to build deb package for systemd
  - I tested on Debian 8 and Ubuntu 16

# This p-r does not introduce:

- Uploading the systemd supported-deb package
- systemd-supported deb package for stage and kcps

I'll do them in another p-r.

PR #368, opened 2017-03-28
Change package License to Apache 2.0

astj:
Discussed on #360.
https://github.com/mackerelio/mackerel-agent/pull/360#discussion_r107351282
https://github.com/mackerelio/mackerel-agent/pull/360#discussion_r107351307

Since mackerel-agent itself is published under Apache License 2.0 and we don't add copyrighted stuff to package releases, we can release the packages under Apache License 2.0 too.
Currently they're released under Commercial License, but it's maybe because at first we don't publish mackerel-agent's source codes.

And also I updated copyright terms in the packages. We, Hatena should be referred as `Hatena Co., Ltd`.

### DIFF
--- a/packaging/deb-systemd/debian/copyright
+++ b/packaging/deb-systemd/debian/copyright
@@ -1,6 +1,6 @@
 This work was packaged for Debian by:
 
-    y_uuki <y_uuki@hatena.ne.jp> on Tue, 11 Mar 2014 11:27:51 +0900
+    astj <astj@hatena.ne.jp> on Tue, 28 Mar 2017 11:09:51 +0900
 
 It was downloaded from:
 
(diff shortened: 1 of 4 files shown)


packaging/rpm/mackerel-agent-systemd.spec
```diff
@@ -6,14 +6,14 @@
 Name:      mackerel-agent
 Version:   %{_version}
 Release:   1%{?dist}
-License:   Commercial
+License:   ASL 2.0
```
astj: https://fedoraproject.org/wiki/Packaging:LicensingGuidelines

> The License: field must be filled with the appropriate license Short License identifier(s) from the "Good License" tables on the Fedora Licensing page.

And Short License Identifier for Apache License 2.0 is `ASL 2.0`, according to the table.
https://fedoraproject.org/wiki/Licensing:Main

packaging/deb-systemd/debian/copyright
```diff
@@ -1,19 +1,19 @@
...
 
 Copyright:
 
-    <Copyright (C) 2014 Hatena>
+    <Copyright (C) 2017 Hatena Co., Ltd.>
 
 License:
 
-    Commercial
+    Apache-2.0
```
astj: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/#license-specification

itchyny:
LGTM

astj:
Thanks!

PR #369, opened 2017-03-28
Release systemd deb packages to github releases

astj:
Release #360 to github releases. Also implemented make tasks for staging/kcps systemd debs
(Related: #347)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -170,4 +170,4 @@ clean:
 
 generate: commands_gen.go
 
-.PHONY: test build run deps clean lint crossbuild cover rpm deb tgz generate crossbuild-package crossbuild-package-kcps crossbuild-package-stage rpm-v1 rpm-v2 rpm-stage rpm-stage-v1 rpm-stage-v2 rpm-kcps-v1 rpm-kcps-v2 deb-v1 deb-v2 deb-kcps-v1 deb-kcps-v2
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -156,4 +170,4 @@ clean:
 
 generate: commands_gen.go
 
-.PHONY: test build run deps clean lint crossbuild cover rpm deb tgz generate crossbuild-package crossbuild-package-kcps crossbuild-package-stage rpm-v1 rpm-v2 rpm-stage rpm-stage-v1 rpm-stage-v2 rpm-kcps-v1 rpm-kcps-v2 deb-systemd
+.PHONY: test build run deps clean lint crossbuild cover rpm deb tgz generate crossbuild-package crossbuild-package-kcps crossbuild-package-stage rpm-v1 rpm-v2 rpm-stage rpm-stage-v1 rpm-stage-v2 rpm-kcps-v1 rpm-kcps-v2 deb-v1 deb-v2 deb-kcps-v1 deb-kcps-v2
```
itchyny: deb-stage-v[12] should be added here as well IMO

itchyny: Also, I think I would add deb-kcps, deb-stage here too, but am I missing something?

astj: I just forgot to add..
(We can omit .PHONY for task XXX unless a file with the name XXX exists on the same directory, but it's better to add to avoid confusion.)

itchyny:
LGTM

astj:
Thanks!

PR #370, opened 2017-03-30
Change systemd deb package architecture to amd64

astj:
Currently systemd deb package released as `all` architecture, but it includes amd64 binary.
I'd like to change package architecture to amd64 so that package architecture will be the same as including binary.


itchyny:
LGTM

astj:
Thanks!

PR #371, opened 2017-03-30
[nit] Add pointer for full license text to debian package license information

astj:
related: #368 
This will reduce lintian errors on building deb packages: `E: mackerel-agent: copyright-should-refer-to-common-license-file-for-apache-2`.
In fact there're still many errors and warnings, but reducing error is better than doing nothing...

Ref: https://lintian.debian.org/tags/copyright-should-refer-to-common-license-file-for-apache-2.html


itchyny:
LGTM!

PR #375, opened 2017-04-11
Adjust config package

Songmu:
- Do not use "mackerel-agent" immediate value in config_linux.go, use `getAgentName()` instead.
- Commonize config_linux.go, config_freebsd.go and config_netbsd.go
- declare `DefaultConfig` in config.go and assign default values into it in `init()` function.


itchyny:
LGTM!

PR #380, opened 2017-04-26
Output error logs of mackerel-agent as warning log of windows event log

Songmu:
Since eventlog is a system-wide log, error logs of mackerel-agent are handled as warning level in the eventlog


mattn:
I agree with this change.

PR #382, opened 2017-04-28
use DiskReadsPerSec/DiskWritesPerSec instead of DiskReadBytesPersec/DiskWriteBytesPersec

mattn:
disk

should be count not bytes.


daiksy:
LGTM!

PR #384, opened 2017-05-05
rename command.Context to command.App

Songmu:
Because it is confusing with context.Context.


astj:
LGTM. (I don't know why coverage decreased by this change...)

PR #387, opened 2017-05-10
Add `prevent_alert_auto_close` option for check plugins

mechairoi:
If `disable_auto_close` option is true, alerts of the check plugin are no longer closed automatically.
This option make mackerel-agent not report `OK` status of the plugin to the server.

`disable_auto_close = true` option is not available in combination with `max_check_attempts > 1` option.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -87,6 +87,7 @@ type PluginConfig struct {
 	ExecutionInterval    *int32  `toml:"execution_interval"`
 	MaxCheckAttempts     *int32  `toml:"max_check_attempts"`
 	CustomIdentifier     *string `toml:"custom_identifier"`
+	DisableAutoClose     *bool   `toml:"disable_auto_close"`
 }
 
 // MetricPlugin represents the configuration of a metric plugin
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -136,21 +137,28 @@ type CheckPlugin struct {
...
 		Command:              pconf.Command,
 		CommandArgs:          pconf.CommandArgs,
 		User:                 pconf.User,
 		NotificationInterval: pconf.NotificationInterval,
 		CheckInterval:        pconf.CheckInterval,
 		MaxCheckAttempts:     pconf.MaxCheckAttempts,
-	}, nil
+	}
+	if pconf.DisableAutoClose == nil {
+		plugin.DisableAutoClose = false
```
astj: [nit]
You can omit explicitly `false` assignment, since it's treated as `false` when it's not assigned.

astj: Using `bool` (not `*bool`) in `pconf.DisableAutoClose` can simplify this assignment?

mattn: If you expect false in default, EnableAutoClose is better. (This is not strong opinion)

mechairoi: I expect `EnableAutoClose = true`(`DisableAutoClose = false`) in default for backward compatibility.

itchyny: BTW user cannot tell _close what?_. Some suggestions: `keep_alert_open`, `prevent_alert_auto_close`.

stefafafan: I personally like `prevent_alert_auto_close` , it seems simple and clear.

astj:
LGTM!

PR #391, opened 2017-06-05
remove version package and adjust internal dependencies

Songmu:
- follow the convention of version declaration, like other products (mackerel-agent-plugins, go-check-plugins and mkr)
- there is something strange that the version package is independent
- API client (mackerel/api.go) should not depend on the internal packages like version
  - I want to separate it and merge to mackerelio/mackerel-client-go in the future


Makefile
```diff
@@ -1,23 +1,16 @@
 MACKEREL_AGENT_NAME ?= "mackerel-agent"
 MACKEREL_API_BASE ?= "https://mackerel.io"
-MACKEREL_AGENT_VERSION ?= $(shell git describe --tags --abbrev=0 | sed 's/^v//' | sed 's/[-+].*$$//')
+VERSION = $(shell git log --merges --oneline | perl -ne 'if(m/^.+Merge pull request \#[0-9]+ from .+\/bump-version-([0-9\.]+)/){print $$1;exit}')
```
astj: Hmm, this seems little bit tricky... Can we use git tags instead?

Songmu: I know that. So, I want to declare version explicitly here in future (next pull request).
We can't use git tag, because release build is proceeded before git tagged.
(Build the artifacts, git tag it, and put it in github releases.)

mackerel/api.go
```diff
@@ -73,13 +73,18 @@ func (api *API) urlFor(path string, query string) *url.URL {
...
+		return api.UA
+	}
+	return "mackerel-agent/0.0.0"
+}
+
 var apiRequestTimeout = 30 * time.Second
 
 func (api *API) do(req *http.Request) (resp *http.Response, err error) {
 	req.Header.Add("X-Api-Key", api.APIKey)
-	req.Header.Add("X-Agent-Version", version.VERSION)
```
astj: We don't use `X-Agent-Version` and `X-Revision` header currently, so maybe it's okay. maybe……

itchyny:
LGTM

PR #395, opened 2017-06-23
Migrate to mackerelio/golib/logging

astj:
ref: https://github.com/mackerelio/golib/blob/master/logging/logger.go

I'd like to remove mackerel-agent/logging in near future, but we cannot do it now because plugins are still using it.


Songmu:
Please replace the plugin side first. The situation where logger and logging coexist is likely to cause confusion.

Songmu:
Merge it after removing mackerel-agent/logging

Songmu:
LGTM

astj:
Thanks!

PR #398, opened 2017-07-12
Improve the EC2 Instance check

stefafafan:
The isEC2 check always calls the metadata API, and does not retry. Instead, for EC2 linux instances we can first check the UUID and only call the API if the UUID starts with `ec2` . Also, perhaps retry when that is the case.

ref. http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/identify_ec2_instances.html

### DIFF
--- a/spec/cloud.go
+++ b/spec/cloud.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


itchyny:
LGTM

stefafafan:
Thank you!

PR #401, opened 2017-07-20
Adjust isEC2 check 

stefafafan:
- Made the retry interval longer (from 200ms to 2000ms)
- Made it not retry calling the Metadata API for non Linux instances

### DIFF
--- a/spec/isec2.go
+++ b/spec/isec2.go
@@ -13,5 +13,4 @@ func isEC2() bool {
 	defer resp.Body.Close()
 
 	return resp.StatusCode == 200
-
 }
--- a/spec/isec2.go
+++ b/spec/isec2.go
@@ -1,6 +1,9 @@
+// +build !linux
+
 package spec
 
-func isEC2ForNonLinux() bool {
+// For instances other than Linux, call Metadata API only once.
+func isEC2() bool {
 	cl := httpCli()
 	// '/ami-id` is probably an AWS specific URL
 	resp, err := cl.Get(ec2BaseURL.String() + "/ami-id")
--- a/spec/isec2.go
+++ b/spec/isec2.go
@@ -1,0 +1,14 @@
+package spec
+
+func isEC2ForNonLinux() bool {
+	cl := httpCli()
+	// '/ami-id` is probably an AWS specific URL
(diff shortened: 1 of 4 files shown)


itchyny:
LGTM

stefafafan:
@itchyny I'm sorry I made some more edits, in order to avoid calling the metadata API too many times I stopped the retry for non-Linux instances.

Songmu:
Since the difference got bigger, could you separate the implementation of `isEC2` into files for linux and other environment?
e.g. `isec2_linux.go` and `isec2.go`.

stefafafan:
I've moved the implementations to separate files.

spec/isec2.go
```diff
@@ -0,0 +1,17 @@
...
+	cl := httpCli()
+	// '/ami-id` is probably an AWS specific URL
+	resp, err := cl.Get(ec2BaseURL.String() + "/ami-id")
+	if err != nil {
+		return false
+	}
+	defer resp.Body.Close()
+
+	return resp.StatusCode == 200
+
```
itchyny: [nit] delete this line

spec/isec2_linux.go
```diff
@@ -0,0 +1,41 @@
...
+
+// If the OS is Linux, check /sys/hypervisor/uuid file first. If UUID seems to be EC2-ish, call the metadata API (up to 3 times).
+// ref. http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/identify_ec2_instances.html
+func isEC2() bool {
+	data, err := ioutil.ReadFile("/sys/hypervisor/uuid")
+	if err != nil {
+		// Probably not EC2.
+		return false
+	}
+	// Probably not EC2.
```
itchyny: What does `Probably` mean? Do you mean this method can return false even if in EC2?

stefafafan: Ah, nice catch. 
- If the UUID has prefix of "ec2", it is probably EC2
- If the UUID does not have a prefix of "ec2", it is not EC2

Thus, this comment seems incorrect as you mentioned.

itchyny:
LGTM

stefafafan:
Thanks for the review!

PR #404, opened 2017-07-27
Fix the Azure VM check

stefafafan:
It seems like I only defined the Azure VM metadata related methods, but never called isAzure..


stefafafan:
Thanks!

PR #408, opened 2017-08-25
Show CPU/SoC model name on Linux/MIPS

hnw:
When using mackerel-agent on Linux/MIPS, the CPU/SoC model name is "undefined" in the Mackerel console. 

image

On Linux/MIPS, the output of `/proc/cpuinfo` is as follows:

```
# cat /proc/cpuinfo
system type		: Atheros AR7242 rev 1
machine			: Buffalo BHR-4GRV
processor		: 0
cpu model		: MIPS 24Kc V7.4
BogoMIPS		: 265.42
wait instruction	: yes
microsecond timers	: yes
tlb_entries		: 16
extra interrupt vector	: yes
hardware watchpoint	: yes, count: 4, address/irw mask: [0x0ffc, 0x0ffc, 0x0ffb, 0x0ffb]
isa			: mips1 mips2 mips32r1 mips32r2
ASEs implemented	: mips16
shadow register sets	: 1
kscratch registers	: 0
package			: 0
core			: 0
VCED exceptions		: not available
VCEI exceptions		: not available
```

This is a slightly different result than the Linux/x86_64 architecture. The CPU/SoC model name is displayed as "system type", however mackerel-agent doesn't support it. 

This behavior can also be confirmed by checking the linux source code: [arch/mips/kernel/proc.c](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/arch/mips/kernel/proc.c?h=v4.13-rc6#n53) .

### DIFF
--- a/spec/linux/cpu.go
+++ b/spec/linux/cpu.go
@@ -44,7 +44,7 @@ func (g *CPUGenerator) generate(file io.Reader) (interface{}, error) {
 				cur["model_name"] = modelName
 			}
 			results = append(results, cur)
-		case "Processor":
+		case "Processor", "system type":
 			modelName = val
 		case "vendor_id", "model", "stepping", "physical id", "core id", "model name", "cache size":
 			cur[strings.Replace(key, " ", "_", -1)] = val


itchyny:
That host page shows the value of `model_name` field. And it looks like you are changing the wrong place.
```diff
diff --git a/spec/linux/cpu.go b/spec/linux/cpu.go
index 0c15175..dfb19ed 100644
--- a/spec/linux/cpu.go
+++ b/spec/linux/cpu.go
@@ -46,7 +46,7 @@ func (g *CPUGenerator) generate(file io.Reader) (interface{}, error) {
 			results = append(results, cur)
 		case "Processor":
 			modelName = val
-		case "vendor_id", "model", "stepping", "physical id", "core id", "model name", "cache size":
+		case "vendor_id", "model", "stepping", "physical id", "core id", "model name", "system type", "cache size":
 			cur[strings.Replace(key, " ", "_", -1)] = val
 		case "cpu family":
 			cur["family"] = val
```
We Mackerel developer team will survey if that information is the right thing to show as the model name, but maybe we can fixing the host page to show the `system_type` field as well. Adding a test is recommended.

itchyny:
Another possible fix.
```diff
diff --git a/spec/linux/cpu.go b/spec/linux/cpu.go
index 0c15175..35feb41 100644
--- a/spec/linux/cpu.go
+++ b/spec/linux/cpu.go
@@ -48,6 +48,8 @@ func (g *CPUGenerator) generate(file io.Reader) (interface{}, error) {
 			modelName = val
 		case "vendor_id", "model", "stepping", "physical id", "core id", "model name", "cache size":
 			cur[strings.Replace(key, " ", "_", -1)] = val
+		case "system type":
+			cur["model name"] = val
 		case "cpu family":
 			cur["family"] = val
 		case "cpu cores":
```
By the way, the `cpu model: MIPS 24Kc V7.4` is important or not?

hnw:
Oh, you are right. I pushed incorrect commit.

itchyny:
LGTM!

hnw:
My understanding is that `cpu model` means an architecture name or an instruction set name developed by MIPS Technologies. `system type` shows compatible CPU/SoC name which other licensed vendor implements.

PR #413, opened 2017-09-19
[WIP] introduce secondary API client

Songmu:


Songmu:
I close it because it is too much.
It is enough to change the API domain once.

PR #416, opened 2017-09-28
Filter plugin metrics value by include_pattern and exclude_pattern option

astj:
When `ruby ./dice.rb` generates two metrics `dice6` and `dice20`, following plugin config now only sends `dice6` metrics to Mackerel.

```toml
[plugin.metrics.dice]
command = "ruby ./dice.rb"
include_pattern = '^dice6'
```

And `exclude_pattern` will exclude metrics if the metric name matches `exclude_pattern`.


mechairoi:
LGTM!

astj:
Thank you!

PR #417, opened 2017-10-02
Use new API BaseURL

astj:
Our announcement:
- En: https://mackerel.io/blog/entry/announcement/20170921
- Ja: https://mackerel.io/ja/blog/entry/announcement/20170921 

Now we use `api.mackerelio.com` instead of `mackerel.io` as the base URL for API requests.
`https://api.mackerelio.com/api/...` should behave as same as `https://mackerel.io/api/...`. 

### Note

This change only affects to API requests.  Mackerel console `https://mackerel.io/` is not affected. (We'll update other our projects, like `mackerel-client-go` and `mkr` also.)
And `https://mackerel.io/api/...` is still available, therefore immediate migration is not necessary.


itchyny:
LGTM

PR #420, opened 2017-10-13
Do not send too many reports in one API request.

astj:
Currently all check monitoring reports are sent to Mackerel in one API request, but it may contain too many reports in one API request, if the host so many check monitoring.


astj:
Thanks!

PR #422, opened 2017-10-15
Use go-osstat library on darwin

itchyny:
## Why?
In the current implementation of mackerel-agent, codes for collecting metrics are tightly coupled within the codes for the agent.
I decided to split the codes for metrics into an external library [mackerelio/go-osstat](https://github.com/mackerelio/go-osstat).
By splitting the code, it's expected to get easier to improve the performance, add more tests and reduce environment-depended codes from mackerel-agent.

As the first step, I migrated the code for darwin to go-osstat in this pull request.

#### Screenshot
screen shot 2017-10-15 at 8 59 19

#### Before: `sysctl -n vm.loadavg`.
```
 $ sysctl -n vm.loadavg
{ 2.27 2.46 2.49 }
```

<s>#### After: `getloadavg(3)`.
```
GETLOADAVG(3)            BSD Library Functions Manual            GETLOADAVG(3)

NAME
     getloadavg -- get system load averages

LIBRARY
     Standard C Library (libc, -lc)

SYNOPSIS
     #include <stdlib.h>

     int
     getloadavg(double loadavg[], int nelem);
```
</s>

#### After: `sysctl(3) vm.loadavg`
https://gist.github.com/itchyny/f06bd3df881b3d013787e4d224aa7b24
(cgo is unavailable because we use crossbuild on TravisCI)

#### Difference
- improvement in accuracy
  - before: two digits after the decimal point
  - after: double precision floating point
- improvement in performance

<s>### CPU usage
#### Screenshot
screen shot 2017-10-15 at 8 58 53

#### Before: `iostat -n0 -c2`
```
 $ iostat -n0 -c2
      cpu    load average
 us sy id   1m   5m   15m
 11  6 83  2.37 2.47 2.49
 17  7 76  2.26 2.45 2.48
```

#### After: `host_statistics(.., HOST_CPU_LOAD_INFO, ..)`
https://gist.github.com/itchyny/f9a759feebe9415443b095ae14074288
https://developer.apple.com/documentation/kernel/1502546-host_statistics?language=objc
https://opensource.apple.com/source/system_cmds/system_cmds-433/iostat.tproj/iostat.c.auto.html

#### Difference
- improvement in accuracy
  - before: rounded into integer number, difference of statistics counter in one second
  - after: double precision floating point, difference in 60 seconds (`metricsInterval`)
- improvement in performance
</s>

⚠️  Edit: I decided not to use go-osstat for darwin because cpu tick counters are unavailable without cgo (we cross-build on Linux).

#### Screenshot
screen shot 2017-10-15 at 8 59 05

#### Before: `vm_stat`
```
 $ vm_stat
Mach Virtual Memory Statistics: (page size of 4096 bytes)
Pages free:                               20161.
Pages active:                            967103.
Pages inactive:                          934255.
Pages speculative:                        10709.
Pages throttled:                              0.
Pages wired down:                       1673077.
Pages purgeable:                          39301.
...
```

#### After: `vm_stat`
Same as before.

### Difference
- None

#### Before: `sysctl vm.swapusage`
 $ sysctl vm.swapusage
vm.swapusage: total = 5120.00M  used = 3693.25M  free = 1426.75M  (encrypted)

#### After: `sysctl(3) vm.swapusage`
https://gist.github.com/itchyny/3c4b56c9168b214643d8990ca85ba7d8

#### Difference
- improvement in performance

#### Screenshot
screen shot 2017-10-15 at 8 59 12

#### Before: `netstat -bni`
```
Name  Mtu   Network       Address            Ipkts Ierrs     Ibytes    Opkts Oerrs     Obytes  Coll
lo0   16384 <Link#1>                       1758927     0  601295906  1758927     0  601295906     0
lo0   16384 127           127.0.0.1        1758927     -  601295906  1758927     -  601295906     -
lo0   16384 ::1/128     ::1                1758927     -  601295906  1758927     -  601295906     -
lo0   16384 fe80::1%lo0 fe80:1::1          1758927     -  601295906  1758927     -  601295906     -
gif0* 1280  <Link#2>                             0     0          0        0     0          0     0
stf0* 1280  <Link#3>                             0     0          0        0     0          0     0
en0   1500  <Link#4>    a4:5e:60:aa:aa:aa 23300388     0 18096041919 17727990     0 8602191509     0
en0   1500  fe80::222:a fe80:4::222:aaaa: 23300388     - 18096041919 17727990     - 8602191509     -
en0   1500  192.168.105   192.168.105.102 23300388     - 18096041919 17727990     - 8602191509     -
en1   1500  <Link#5>    4a:00:00:aa:aa:aa        0     0          0        0     0          0     0
en2   1500  <Link#6>    4a:00:00:aa:aa:aa        0     0          0        0     0          0     0
```

#### After: `netstat -bni`
Same as before.

#### Difference
- improvement in performance (less number of regexp to parse the output of the command)


itchyny:
Still have problems for CPU statistics on non-cgo environment.

itchyny:
I decided to revert cf53f71346c77157cdcebae36b497c1bf30e1f14 because I could not find a way (command or sysctl) to get the cpu tick counter without cgo. We cross-build darwin binaries on Linux of TravisCI so cgo is unavailable. I note that the cpu usage on darwin is not accurate (difference between one second, rounded to integers).

command/command_darwin.go
```diff
@@ -26,7 +26,6 @@ func metricsGenerators(conf *config.Config) []metrics.Generator {
 		&metricsDarwin.Loadavg5Generator{},
 		&metricsDarwin.CPUUsageGenerator{},
 		&metricsDarwin.MemoryGenerator{},
-		&metricsDarwin.SwapGenerator{},
```
astj: [memo]
Now Swap metrics are retrieved in MemotyGenerator

metrics/darwin/interface.go
```diff
@@ -59,73 +45,26 @@ func (g *InterfaceGenerator) Generate() (metrics.Values, error) {
...
-	}
-	// results (eth0) sample
-	/** [%!s(*metrics.Value=&{interface.eth0.rxBytes 6.6074069281e+10})
-	     %!s(*metrics.Value=&{interface.eth0.txBytes 9.180531994e+09})
-	    ]
-	**/
-	if hasNonZeroValue {
-		return metrics.Values(results), nil
+	if len(networks) == 0 {
+		return nil, nil
```
astj: Does it make difference between returning empty map and returning nil?

itchyny: No, nothing crashes.
```go
gore> x := func() map[string]uint64 { return nil }()
map[string]uint64(nil)
gore> x
map[string]uint64(nil)
gore> for _, name := range x { println(name) }
gore> x["interface.foo.rxBytes"]
0x0
gore> currValue, ok := x["interface.foo.rxBytes"]
0x0
false
```

itchyny:
Thanks for review.

PR #424, opened 2017-10-16
Change log level in device generator if /sys/block does not exist

### LINKED ISSUE #421: spec.block_device error to use mackerel agent on Heroku
Hi,

I installed mackerel-agent with [this article](https://mackerel.io/ja/docs/entry/advanced/monitoring-heroku) on Heroku,
then there are some errors like below every hour:

```
block_device.go:29: ERROR <spec.block_device> Failed (skip this spec): open /sys/block: no such file or directory
```

I use only 1 dyno, and Django is working on Heroku.
How do I do to stop this error ?

Thanks.

itchyny:
Fixes #421. Some the directory is unavailable on some PaaS.


itchyny:
I think this p/r is not a good solution... better to change the log level to warning?

itchyny:
Thank you!

PR #429, opened 2017-10-18
Ensure returned value of retrieveAzureVMMetadata is not null

astj:
ref: https://github.com/mackerelio/mackerel-agent/blob/b346699b247364717240c9a35bbfa10748e67ce6/spec/cloud.go#L282-L284

When first `retrieveAzureVMMetadata` fails but second succeeds. second `retrieveAzureVMMetadata` tries to access `metadata`, which became `nil` by first `retrieveAzureVMMetadata`.

With this change, returned values of `retrieveAzureVMMetadata` is guaranteed not be `nil` (unless `metadataMap` parameter is `nil` initially), so that parameter of `retrieveAzureVMMetadata` in L284 will be always non-nil.


itchyny:
LGTM

astj:
Thank you.

PR #439, opened 2017-11-05
[nit] Print timed out command itself

syou6162:
When the command ran by mackerel-agent timed out, mackerel-agent print following error message to `mackerel-agent.log`.

```
2017/11/06 06:29:42 ERROR <util> RunCommand error command: []string, error: command timed out
```

We cannot know which command actually timed out. WIth this pull request, mackerel-agent prints the command when it timed out.

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -5,7 +5,6 @@ import (
 	"os"
 	"os/exec"
 	"runtime"
-	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


itchyny:
Considering the readability joining by spaces works nice, but I think it's better to use `%v` to reconstruct the actual command to be invoked. What do you think?

itchyny:
LGTM

PR #441, opened 2017-11-09
Commonize interface generators for Linux, Darwin and add support for BSD systems

itchyny:
I refactored interface generators, commonize the generators. I also introduced this generator for BSD systems.


astj:
LGTM.

itchyny:
Thank you.

PR #442, opened 2017-11-14
Refine metrics collector

mechairoi:
In a specific environment, metrics not sent each minutes.

- Reduce gap between agent.MetricsResult.Created and the time of metriss collected.
  - When metrics collectors spent long time, I think there is possibilities that the gap is 10 or more minutes.
-  Log durations that each metrics generator spent 
  - To detect slow metrics generator.

### DIFF
--- a/agent/generator.go
+++ b/agent/generator.go
@@ -2,6 +2,7 @@ package agent
 
 import (
 	"sync"
+	"time"
 
 	"github.com/mackerelio/golib/logging"
 	"github.com/mackerelio/mackerel-agent/metrics"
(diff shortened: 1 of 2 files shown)


agent/generator.go
```diff
@@ -39,7 +40,11 @@ func generateValues(generators []metrics.Generator) []*metrics.ValuesCustomIdent
 					wg.Done()
 				}()
 
+				startedAt := time.Now()
 				values, err := g.Generate()
+				if seconds := (time.Now().Sub(startedAt) / time.Second); seconds > 120 {
+					logger.Warningf("%T.Generate() take a long time (%d seconds)", g, (time.Now().Sub(startedAt) / time.Second))
```
astj: You can re-use `seconds` here

PR #443, opened 2017-11-14
Improve plugin command parsing error message

itchyny:
This pull request improves the error message on parsing command. When `command` for (metric|check|metadata) plugin is not configured correctly, agent should report errors which plugin configuration is not correct.
```
failed to test config: failed to load the config file: plugin.metrics.mysql: failed to parse plugin command. A configuration value of `command` should be string or string slice, but
 int64
```
The message contains `plugin.metrics.mysql` so that user can configure which plugin configuration is invalid.

Example invalid conf:
```
apikey = "abcde"

[plugin.checks.ok]
command = 'echo ok'

[plugin.metrics.mysql]
command = 100
user = "mysql"
```

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -12,6 +12,7 @@ import (
 	"github.com/BurntSushi/toml"
 	"github.com/mackerelio/golib/logging"
 	"github.com/mackerelio/mackerel-agent/util"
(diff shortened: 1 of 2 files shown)


itchyny:
Thanks!

PR #447, opened 2017-11-30
Add an option to declare cloud platform explicitly

astj:
In mackerel-agent.conf:
```
cloud_platform = "ec2"
```

available values are following:
- ec2, azurevm, gce
- auto: default behavior. will determine platform using `isXXX()`
- none: does not use CloudGenerator

### DIFF
--- a/spec/cloud_test.go
+++ b/spec/cloud_test.go
@@ -177,10 +177,11 @@ func TestGCEGenerate(t *testing.T) {
 }
 
 func TestSuggestCloudGenerator(t *testing.T) {
-	// both of ec2BaseURL and gceMetaURL are unreachable
+	// All Cloud meta URLs are unreachable
 	unreachableURL, _ := url.Parse("http://unreachable.localhost")
 	ec2BaseURL = unreachableURL
 	gceMetaURL = unreachableURL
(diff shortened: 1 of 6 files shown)


spec/cloud_test.go
```diff
@@ -218,4 +225,97 @@ func TestSuggestCloudGenerator(t *testing.T) {
...
+}
+
+func TestSuggestCloudGenerator_CloudPlatformSpecified(t *testing.T) {
+	// All Cloud meta URLs are unreachable
+	unreachableURL, _ := url.Parse("http://unreachable.localhost")
+	ec2BaseURL = unreachableURL
+	gceMetaURL = unreachableURL
+	azureVMBaseURL = unreachableURL
+
+	func() {
```
itchyny: [nit] Lexical scoping of blocks instead of function should work here.

itchyny:
LGTM!

spec/cloud_test.go
astj: https://github.com/mackerelio/mackerel-agent/pull/447/commits/6d9f5d2a098ffaf4fb8907e95409eb967cce043c

PR #452, opened 2017-12-18
Fix network interface spec collector on Windows

itchyny:
This pull request fixes network interface spec on Windows.

### DIFF
--- a/spec/windows/interface.go
+++ b/spec/windows/interface.go
@@ -59,6 +59,8 @@ func (g *InterfaceGenerator) Generate() ([]spec.NetInterface, error) {
 			switch v := addr.(type) {
 			case *net.IPAddr:
 				ip = v.IP
+			case *net.IPNet:
+				ip = v.IP
 			}
 			if ip == nil || ip.IsLoopback() {
 				continue


spec/windows/interface.go
```diff
@@ -59,6 +59,8 @@ func (g *InterfaceGenerator) Generate() ([]spec.NetInterface, error) {
 			switch v := addr.(type) {
 			case *net.IPAddr:
 				ip = v.IP
+			case *net.IPNet:
```
astj: Should we consider range of v.Mask?

astj:
I don't know what subnet of an interface mean, and whether v.IP is proper representation of the interface's IP address...

itchyny:
I think subnet mask is commonly used in local network. The network interface spec collector for Linux also ignores subnet masks ([ref](https://github.com/mackerelio/mackerel-agent/blob/fbbe8d3e086fddaabf5c163165e977251551ea70/spec/linux/interface.go#L60)). I'm not sure users want to see subnet masks or network addresses, but it's reasonable to have the same implementation on each architectures.

spec/windows/interface.go
mattn: How about replace this to `net.Parse(addr.String())` ?

astj:
Okay, I understand.

spec/windows/interface.go
itchyny: `net.ParseIP` does not accept addresses with subnet mask.

itchyny:
Thank you.

PR #453, opened 2017-12-19
fix the retry of check reports

hayajo:
* Retry until report succeeds
* Buffer reports for do not block checking
* Add delay for do not report many times in a short time

### DIFF
--- a/command/command.go
+++ b/command/command.go
@@ -454,8 +454,9 @@ func runChecker(checker *checks.Checker, checkReportCh chan *checks.Report, repo
 // which run for each checker commands and one for HTTP POSTing
 // the reports to Mackerel API.
 func runCheckersLoop(app *App, termCheckerCh <-chan struct{}, quit <-chan struct{}) {
-	checkReportCh := make(chan *checks.Report)
-	reportImmediateCh := make(chan struct{})
+	// Do not block checking.
+	checkReportCh := make(chan *checks.Report, app.Config.Connection.ReportCheckBufferSize*len(app.Agent.Checkers))
+	reportImmediateCh := make(chan struct{}, app.Config.Connection.ReportCheckBufferSize*len(app.Agent.Checkers))
 
 	for _, checker := range app.Agent.Checkers {
 		go runChecker(checker, checkReportCh, reportImmediateCh, quit)
@@ -478,6 +479,7 @@ func runCheckersLoop(app *App, termCheckerCh <-chan struct{}, quit <-chan struct
 			select {
 			case report := <-checkReportCh:
 				reports = append(reports, report)
+			case <-reportImmediateCh: // drain all
 			default:
 				break DrainCheckReport
 			}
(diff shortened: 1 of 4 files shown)


command/command.go
```diff
@@ -478,6 +479,7 @@ func runCheckersLoop(app *App, termCheckerCh <-chan struct{}, quit <-chan struct
 			select {
 			case report := <-checkReportCh:
 				reports = append(reports, report)
+			case <-reportImmediateCh: // drain all
```
astj: [q]
Since we already received from `reportImmediateCh` in L472, `reportImmediateCh` should have been inserted (at least) twice to receive  here?

config/config.go
```diff
@@ -44,6 +44,10 @@ var defaultConnectionConfig = ConnectionConfig{
 	PostMetricsRetryDelaySeconds:   60,     // Wait a minute before retrying metric value posts
 	PostMetricsRetryMax:            60,     // Retry up to 60 times (30s * 60 = 30min)
 	PostMetricsBufferSize:          6 * 60, // Keep metric values of 6 hours span in the queue
+	ReportCheckDelaySeconds:        1,      // Wait a second before reporting the next check
+	ReportCheckDelaySecondsMax:     30,     // Wait 30 seconds before reporting the next check when many reports in queue
```
astj: imo This delay can be shorter, imo. (just a few seconds seems enough)

command/command.go
hayajo: reportImmediateCh has a buffer to avoid blocking checks.
Since it is unnecessary except the first one, drain all.

command/command.go
```diff
@@ -489,33 +491,42 @@ func runCheckersLoop(app *App, termCheckerCh <-chan struct{}, quit <-chan struct
 
 		// Do not report too many reports at once.
 		const checkReportMaxSize = 10
+
+		// Do not report many times in a short time.
+		reportCheckDelay := app.Config.Connection.ReportCheckDelaySeconds
+		// Extend the delay when there are lots of reports
+		if len(reports) > len(app.Agent.Checkers) {
+			delay := app.Config.Connection.ReportCheckDelaySecondsMax
```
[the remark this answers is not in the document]
hayajo: Thank you for your advice. Fix it.

command/command.go
```diff
@@ -489,33 +491,42 @@ func runCheckersLoop(app *App, termCheckerCh <-chan struct{}, quit <-chan struct
...
-	err := app.API.ReportCheckMonitors(app.Host.ID, reports)
-	if err != nil {
+func reportCheckMonitors(app *App, reports []*checks.Report) {
+	for {
+		err := app.API.ReportCheckMonitors(app.Host.ID, reports)
+		if err == nil {
+			break
+		}
 		logger.Errorf("ReportCheckMonitors: %s", err)
+		logger.Debugf("ReportCheckMonitors: Sleep %d seconds before reporting", app.Config.Connection.ReportCheckRetryDelaySeconds)
```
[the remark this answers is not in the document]
hayajo: Thanks. fix it.

PR #457, opened 2017-12-27
Raise an error in mackerel.FindHost when hostId is empty

astj:
This can happen when `/var/lib/mackerel-agent/id` exits but its contents is (somehow) empty.
Currently, `api.FindHost("")` requests `GET /api/v0/hosts/`, and will be redirected to `/api/v0/host` with status code 200 🙀


itchyny:
I think LoadHostID should not return empty string with no error.

astj:
Okay, I'll open another p-r . (#458)

PR #458, opened 2017-12-28
Ignore empty hostid file

astj:
Delivered from #457 
Just ignore empty hostId file, to avoid panic.  In this case a new host will be registered by mackerel-agent.


itchyny:
We can improve the message in doRetire.

itchyny:
LGTM

astj:
Thanks!

PR #460, opened 2017-12-28
Add an option of timeout duration for executing command

taku-k:
## What
This pull-request introduces the option for specifying a duration after which a command execution will be timeout.

## Why
Running a slightly heavy check command will be easily timeout, so users should be able to adjust the timeout duration.

### DIFF
--- a/checks/checker_unix_test.go
+++ b/checks/checker_unix_test.go
@@ -11,15 +11,14 @@ import (
 )
 
 func TestChecker_CheckTimeout(t *testing.T) {
-	origTimeoutDur := util.TimeoutDuration
-	defer func() {
-		util.TimeoutDuration = origTimeoutDur
-	}()
-	util.TimeoutDuration = 1 * time.Second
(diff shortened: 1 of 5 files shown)


Songmu:
That's a good pull request! We wanted to implement it for a long time.

I point out two points.

## Option name

I think `timeout_seconds` would be better.

## naming `context` is confusing

I think it is better not to use context in naming except for the standard context package
in Go world.

How about naming it as `CommandOption` and adding it to the end of the argument?

(In the first place, the package name util is not good itself. There is room for
improvement there.)

taku-k:
Thank you for the review.
I followed your advice.

Songmu:
Thanks for quick fixing.

One more. You should set `timeoutDururation` for action in `buildCheckPlugin`.

taku-k:
I miss it. Thank you.

Songmu:
Thanks! I

PR #461, opened 2017-12-29
Ignore device-mapper in disk I/O metrics on Linux. (Docker creats a lot)

hayajo:
- Disk read/writes metrics on device-mapper virtual devices are not useful
- Docker creates a lot of device-mapper device nodes
- Ignore "dm-" (device-mapper) devices in disk metrics


itchyny:
LGTM!

PR #462, opened 2017-12-29
Ignore veth in network I/O metrics on Linux. (Docker creats a lot)

hayajo:
- network metrics on veth are not useful
- Docker creates a lot of veth interfaces
- Ignore "vethXXX" interfaces in network metrics


itchyny:
LGTM!

PR #463, opened 2017-12-29
Ignore connection configurations in mackerel-agent.conf

itchyny:
Configuration of `connection` in mackerel-agent.conf is deprecated and no longer supported. User should not be aware of these configurations.


Songmu:
LGTM

itchyny:
Thank you.

PR #466, opened 2017-12-31
Adjust appveyor.yml

Songmu:
The latest go version is available on appveyor environment, so use it.
For compatibility, we use 386 architecture of Go.


fand:
LGTM!

PR #470, opened 2018-01-07
cut out `cmdutil` package from `util` and interface adjustment

Songmu:
The interface of `util.RunCommand` changed at #460, but the interface here is cumbersome now, and since the package name `util` is not good at all, cut out into another package `cmdutil` and adjust interface.


cmdutil/cmdutil_test.go
```diff
@@ -1,6 +1,4 @@
-// +build linux darwin freebsd netbsd windows
```
astj: I wonder which platforms were excluded 🤔

Songmu: plan9! 😆

PR #471, opened 2018-01-08
fix error check in TestStart of start_test.go

Ken2mer:
This pull request is to fix the last check error returned from 'start()' in 'L.91: go func()'. 
Changed to send 'err' to 'done' channel, and receive 'err' from 'done' channel, then correctly check error returned from 'start()' in 'L.91: go func()'.

### DIFF
--- a/start_test.go
+++ b/start_test.go
@@ -87,14 +87,14 @@ func TestStart(t *testing.T) {
 	}
 	conf.SaveHostID(hostID)
 	termCh := make(chan struct{})
-	done := make(chan struct{})
+	done := make(chan error)
 	go func() {
 		err = start(conf, termCh)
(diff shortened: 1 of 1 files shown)


start_test.go
```diff
@@ -87,14 +87,14 @@ func TestStart(t *testing.T) {
 	}
 	conf.SaveHostID(hostID)
 	termCh := make(chan struct{})
-	done := make(chan struct{})
+	done := make(chan error)
 	go func() {
 		err = start(conf, termCh)
-		done <- struct{}{}
+		done <- err
```
astj: [nit]
just `done <- start(...)` looks enough?

astj:
Thank you!

PR #472, opened 2018-01-10
use supervisor mode in sysvinit script

Songmu:
- for crash recovery


Songmu:
## TOOD

- debian package
- adjust reload behavior

PR #484, opened 2018-02-13
Improve error handling when executing commands

Songmu:
If an exit code can be retrieved from the error, take it.


hayajo:
LGTM

cmdutil/testdata/stubcmd.go
```diff
@@ -0,0 +1,53 @@
...
+	"INT":  os.Interrupt,
+	"TERM": syscall.SIGTERM,
+}
+
+func main() {
+	var (
+		trap     = flag.String("trap", "", "signals")
+		exit     = flag.Int("exit", 0, "exit status")
+		trapExit = flag.Int("trap-exit", 0, "exit status when trapping signal")
+		sleep    = flag.Float64("sleep", 0, "sleep seconds")
```
mattn: why not int64?

cmdutil/testdata/stubcmd.go
```diff
@@ -0,0 +1,53 @@
...
+			for _ = range c {
+				fmt.Println("signal received")
+				if *trapExit > 0 {
+					os.Exit(*trapExit)
+				}
+			}
+		}()
+	}
+	if *sleep > 0 {
+		time.Sleep(time.Duration(float64(time.Second) * *sleep))
```
mattn: ditto

cmdutil/testdata/stubcmd.go
Songmu: Because, I wanted to specify sleep seconds as a decimal number like `-sleep 0.3`.

mattn: I prefer to use `flag.Duration` in this case. This is easy to specify seconds (-sleep 3s) or hours/minutes (-sleep 2h45m). But this is problem that is likes and dislikes. And this is just nit-picker. :)

Songmu: Oh, I did not know that `flag.Duration` could take `ms`. I Changed to use flag.Duration.

PR #485, opened 2018-02-14
context support in cmdutil

Songmu:
Add `cmdutil.RunCommandContext()` and` cmdutil.RunCommandArgsContext()`.
Just adding them, these functions are not yet used in the code.


itchyny:
LGTM

PR #486, opened 2018-02-28
Enhance diagnostic mode

Songmu:
- `AgentGenerator` now implements `PluginGenerator` interface and send graph definitions of the graphs
- also collect the number of goroutines of the agent itself additionally

### DIFF
--- a/metrics/plugin.go
+++ b/metrics/plugin.go
@@ -181,13 +181,17 @@ func (g *pluginGenerator) loadPluginMeta() error {
 }
 
 func (g *pluginGenerator) makeCreateGraphDefsPayload() []mackerel.CreateGraphDefsPayload {
-	if g.Meta == nil {
+	return makeCreateGraphDefsPayload(g.Meta)
+}
+
+func makeCreateGraphDefsPayload(meta *pluginMeta) []mackerel.CreateGraphDefsPayload {
+	if meta == nil {
 		return nil
 	}
 
 	payloads := []mackerel.CreateGraphDefsPayload{}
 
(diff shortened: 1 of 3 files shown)


metrics/plugin.go
```diff
@@ -230,8 +233,8 @@ func (g *pluginGenerator) collectValues() (Values, error) {
 	for _, line := range strings.Split(stdout, "\n") {
 		// Key, value, timestamp
 		// ex.) tcp.CLOSING 0 1397031808
-		items := delimReg.Split(line, 3)
-		if len(items) != 3 {
+		items := strings.Fields(line)
```
astj: https://golang.org/pkg/strings/#Fields
> Fields splits the string s around each instance of one or more consecutive white space characters, as defined by unicode.IsSpace, returning a slice of substrings of s or an empty slice if s contains only white space.

It seems this satisfies old definitions `\s\t+`.

https://github.com/google/re2/wiki/Syntax
> whitespace (≡ [\t\n\f\r ])

(Btw, In old definitions we could use `\s+` as delimReg?)

mattn: I'm not sure the string possibly have spaces at begin or end of the lien. strings.Fields trim spaces for each items.

```go
package main

import (
	"regexp"
	"strings"
)

func main() {
	s := " a b c "
	println(len(regexp.MustCompile(`\s+`).Split(s, -1))) // 5
	println(len(strings.Fields(s))) // 3
}
```

Songmu: The result from `strings.Fields` is the preferred behavior and we can reduce one regular expression with it, so I adopt this implementation.

PR #488, opened 2018-03-02
Fix EC2 check for KVM based EC2 instance (e.g. c5 instance)

hayajo:
- KVM based EC2 instance is not exists `/sys/hypervisor/uuid`.
- Fix to check `/sys/devices/virtual/dmi/id/product_uuid` when `/sys/hypervisor/uuid` does not exist.

ref. https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/identify_ec2_instances.html

### DIFF
--- a/spec/isec2_linux.go
+++ b/spec/isec2_linux.go
@@ -8,21 +8,32 @@ import (
 	"github.com/Songmu/retry"
 )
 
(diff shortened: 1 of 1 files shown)


Songmu:
LGTM

PR #491, opened 2018-03-11
care `MemAvailable` in collecting metrics around memory on linux

Songmu:
In cases where `MemAvailable` can be acquired or not, we will branch the type of metrics to collect.


itchyny:
LGTM

PR #493, opened 2018-03-19
change the message level from WARNING to INFO when customIdentifier is not registered

hayajo:
change the message level from WARNING to INFO when customIdentifier is not registered at the inital startup.

> ...
> mackerel-agent[3429]: 2018/03/19 15:16:26 WARNING <command> no host was found for the custom identifier: i-...
> ...

### DIFF
--- a/mackerel/api.go
+++ b/mackerel/api.go
@@ -58,6 +58,21 @@ func apiError(code int, msg string) *Error {
 	}
 }
 
+// InfoError represents Error of log level INFO
+type InfoError struct {
+	Message string
+}
+
+func (e *InfoError) Error() string {
+	return e.Message
+}
+
+func infoError(msg string) *InfoError {
+	return &InfoError{
+		Message: msg,
+	}
+}
+
 // NewAPI creates a new instance of API.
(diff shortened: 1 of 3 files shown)


mackerel/api.go
```diff
@@ -170,7 +170,8 @@ func (api *API) FindHostByCustomIdentifier(customIdentifier string) (*Host, erro
 	}
 
 	if len(data.Hosts) == 0 {
-		return nil, fmt.Errorf("no host was found for the custom identifier: %s", customIdentifier)
+		logger.Infof("no host was found for the custom identifier: %s", customIdentifier)
+		return nil, err
```
astj: With this patch `mackerel.API.FindHostByCustomIdentifier` will not return (non-nil) `err` when a host with the CustomIdentifier is not found, but it should return no-nil error in order to the caller will retry the api request.
https://github.com/mackerelio/mackerel-agent/blob/f680106ca00d9292b7d62238991042780fca5053/command/command.go#L61-L64

https://github.com/mackerelio/mackerel-agent/blob/f680106ca00d9292b7d62238991042780fca5053/command/command.go#L41-L50

PR #496, opened 2018-03-23
Support UUID in little-endian format.

hayajo:
see. https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/identify_ec2_instances.html

> Furthermore, for EC2 instances that are not using Amazon Linux, the distribution's implementation of SMBIOS might represent the UUID in little-endian format, therefore the "EC2" characters do not appear at the beginning of the UUID.

### DIFF
--- a/spec/isec2_linux.go
+++ b/spec/isec2_linux.go
@@ -76,9 +76,5 @@ func isEC2UUID(uuid string) bool {
 	var data uint32
 	binary.Read(r, binary.LittleEndian, &data)
 
-	if conds(fmt.Sprintf("%x", data)) {
-		return true
-	}
-
-	return false
+	return conds(fmt.Sprintf("%x", data))
 }
--- a/spec/isec2_linux.go
+++ b/spec/isec2_linux.go
@@ -1,6 +1,10 @@
 package spec
 
 import (
+	"bytes"
+	"encoding/binary"
+	"encoding/hex"
+	"fmt"
 	"io/ioutil"
 	"strings"
 	"time"
@@ -22,7 +26,7 @@ func isEC2() bool {
 		if err != nil {
 			continue
 		}
-		if strings.HasPrefix(string(data), "ec2") || strings.HasPrefix(string(data), "EC2") {
+		if isEC2UUID(string(data)) {
 			looksLikeEC2 = true
 			break
 		}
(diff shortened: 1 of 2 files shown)


spec/isec2_linux.go
```diff
@@ -51,3 +55,34 @@ func isEC2() bool {
...
+	}
+
+	if conds(uuid) {
+		return true
+	}
+
+	// Check as littele endian.
+	// see. https://docs.aws.amazon.com/ja_jp/AWSEC2/latest/UserGuide/identify_ec2_instances.html
+	fields := strings.Split(uuid, "-")
+	src := []byte(fields[0]) // UUID field: time_low(uint32)
```
itchyny: [memo] This is OK, because strings.Split always returns non-empty slice while the separator argument is not empty string.

spec/isec2_linux.go
```diff
@@ -51,3 +55,34 @@ func isEC2() bool {
...
+
+	dst := make([]byte, hex.DecodedLen(len(src)))
+	hex.Decode(dst, src)
```
itchyny: I think `hex.DecodeString` is simpler, what do you think?

spec/isec2_linux_test.go
```diff
@@ -0,0 +1,137 @@
...
+					defer os.Remove(tn)
+				} else {
+					os.Remove(tn)
+					continue
+				}
+
+				tf.Write([]byte("ec2e1916-9099-7caf-fd21-012345abcdef")) // valid EC2 UUID
+				tf.Close()
+			}
+			defer setUUIDFiles(uuidFiles)()
```
itchyny: I don't think it's working as expected. `isEC2` below seems to be called before `uuidFiles` is set in `setUUIDFiles `.

itchyny: Oops, ignore my comment. It's correct while returning a function for defer is difficult.

spec/isec2_linux.go
```diff
@@ -51,3 +55,34 @@ func isEC2() bool {
...
+
+	r := bytes.NewReader(dst)
+	var data uint32
+	binary.Read(r, binary.LittleEndian, &data)
+
+	if conds(fmt.Sprintf("%x", data)) {
+		return true
+	}
+
+	return false
```
itchyny: [nit] just return cond(fmt... seems to work as the same.

spec/isec2_linux.go
hayajo: Definitely.

itchyny:
LGTM!

spec/isec2_linux.go
```diff
@@ -51,3 +55,26 @@ func isEC2() bool {
...
+
+	// Check as littele endian.
+	// see. https://docs.aws.amazon.com/ja_jp/AWSEC2/latest/UserGuide/identify_ec2_instances.html
+	fields := strings.Split(uuid, "-")
+	decoded, _ := hex.DecodeString(fields[0]) // fields[0]: UUID time_low(uint32)
+	r := bytes.NewReader(decoded)
+	var data uint32
+	binary.Read(r, binary.LittleEndian, &data)
+
+	return conds(fmt.Sprintf("%x", data))
```
mattn: uuid string possibly return 0xEC, 0x50 ?

https://play.golang.org/p/zQa6ka44eUp

mattn: Sorry, I understand what you want to fix. https://docs.aws.amazon.com/ja_jp/AWSEC2/latest/UserGuide/identify_ec2_instances.html

PR #501, opened 2018-06-11
improve PATH handling

astj:
- Make PATH not to end with `:`
- Consider path containing `;` in Windows

This is inspired from #499 , but not depending on external libraries.

### DIFF
--- a/init_unix.go
+++ b/init_unix.go
@@ -5,7 +5,11 @@ package main
 import "os"
 
 func init() {
-	os.Setenv("PATH", "/sbin:/usr/sbin:/bin:/usr/bin:"+os.Getenv("PATH"))
+	if path := os.Getenv("PATH"); path != "" {
+		os.Setenv("PATH", "/sbin:/usr/sbin:/bin:/usr/bin:"+path)
+	} else {
+		os.Setenv("PATH", "/sbin:/usr/sbin:/bin:/usr/bin")
+	}
 	// prevent changing outputs of some command, e.g. ifconfig.
 	os.Setenv("LANG", "C")
 	os.Setenv("LC_ALL", "C")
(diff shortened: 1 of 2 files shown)


itchyny:
LGTM

mattn:
I suggest to add `/usr/local/go/bin`

init_windows.go
```diff
@@ -5,11 +5,19 @@ package main
...
 	"path/filepath"
+	"strings"
 )
 
 func init() {
 	if p, err := os.Executable(); err == nil {
-		os.Setenv("PATH", filepath.Dir(p)+
-			string(filepath.ListSeparator)+os.Getenv("PATH"))
+		dir := filepath.Dir(p)
+		if strings.Index(dir, ";") > -1 {
```
Songmu: [nit] How about using `strings.ContainsRune`?

init_unix.go
```diff
@@ -5,7 +5,11 @@ package main
 import "os"
 
 func init() {
```
Songmu: How about the following implementation? It is matter of preference so, you do not necessarily have to apply it.

```go
paths := []string{"/sbin", "/usr/sbin", "/usr/bin"}
if p := os.Getenv("PATH"); p != "" {
   paths = append(paths, p)
}
os.Setenv("PATH", strings.Join(paths, string(filepath.ListSeparator)))
```

mattn:
> I suggest to add /usr/local/go/bin

This is suggestion. So I don't mind even if you don't want my idea. Most user probably install Go into /usr/local/go (if it is not a packages). And this is idea to enable `go run` or `go env` without absolute PATHs.

astj:
@Songmu @mattn Thank you for suggestions. But I'd like to keep them unchanged.

PR #503, opened 2018-06-20
Retry initial API requests with exponential backoff

astj:
Currently agent retries host APIs 20 times with 3 secs interval (= total duration 1 min) while preparing .
With this p-r. agent will retry those APIs 15 times with exponential backoff intervals. Total duration of those retries will be about 6 hours:
```
$ perl -MList::Util -E 'for(0..20) { $sum += List::Util::min(0.5 * (2 ** $_), 15); } say $sum;'
255.5
```


astj:
Needs reconsideration...

PR #507, opened 2018-07-17
Use RunWithEnv instead of os.Setenv to avoid environment variable races

itchyny:
Using os.Setenv here may cause unexpected behavior when collectValues and loadPluginMeta are called concurrently. No issue report is filed yet though (it does not happen luckily or never happen when GOMAXPROCS is set to 1, even though I think we should fix this).


itchyny:
Thank you.

PR #508, opened 2018-07-19
Replace GCE metadata endpoint with absolute FQDN

i2tsuki:
The internal domain is not a TLD. If domain is specified in /etc/resolv.conf, name resolution will be performed as a subdomain of that domain. Therefore, it is necessary to write absolute FQDN in GCE Metadata endpoint.


itchyny:
LGTM!

i2tsuki:
Thanks!!

PR #511, opened 2018-07-25
[windows] Add timeout to WMI query for disk metrics

astj:
wmi query sometimes hangs, therefore add timeout for safety.


astj:
Thanks!

PR #515, opened 2018-08-08
[nit] Refactoring: spec generator returns structs of HostMeta of mackerel-client-go

itchyny:
This pull request refactors spec generators. I prefer using structs to `map[string]interface{}` and let JSON marshalization to the JSON tags. This is a part of refactoring, towards using the client-go library and remove the mackerel directory from this project.

### DIFF
--- a/mackerel/api_test.go
+++ b/mackerel/api_test.go
@@ -8,6 +8,8 @@ import (
 	"net/http/httptest"
 	"reflect"
 	"testing"
+
+	mkr "github.com/mackerelio/mackerel-client-go"
 )
 
 func TestNewAPI(t *testing.T) {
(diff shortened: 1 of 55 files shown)


mackerel/api_test.go
```diff
@@ -8,6 +8,8 @@ import (
 	"net/http/httptest"
 	"reflect"
 	"testing"
+
+	mkr "github.com/mackerelio/mackerel-client-go"
```
astj: [imo]
I don't prefer `mkr` as the alias of mackerel-client-go, since we have a product named `mkr`…

astj:
Though those changes are generally approved (not yet carefully reviewed), but they are not described by  "use HostMeta of mackerel-client-go", and include more changes. Can you check the description and changed?

mackerel/api_test.go
itchyny: I agree but can't help until we have mackerel package here.

astj: 😭

itchyny:
Thank you.

PR #516, opened 2018-08-21
Do HTTP retry on determining cloud platform and suggesting customIdentifier

astj:
In some unlucky situations, HTTP requests to internal metadata endpoint might fail accidentally.
This p-r introduces:
- Retry on SuggestCustomIdentifier itself
    - 1 time => up to 3 times with 2 seconds interval
- Concurrent requests to determine cloud platform
    - (isEC2 => isGCE => isAzure) => (isEC2, isGCE, isAzure)
- Retry on determining cloud platform by HTTP requests
    - 1 time => up to 2 times with 2 seconds interval (exception: EC2 on Linux already retries)

Concurrent requesting `isXXX` is introduced to reduce impact on retry in `isXXX`, which may take up to (3+2+3) seconds for each `isXXX`.

And also, I modified `isEC2` in Linux to avoid data race in tests (happend in  
https://travis-ci.org/mackerelio/mackerel-agent/builds/418496127 ).

### DIFF
--- a/spec/cloud.go
+++ b/spec/cloud.go
@@ -160,25 +160,6 @@ func requestGCEMeta(ctx context.Context) ([]byte, error) {
 	return ioutil.ReadAll(resp.Body)
 }
 
-const maxRetry = 3
-
-func suggestWithRetry(f func() (string, error)) (string, error) {
-	var (
-		retry      int
-		identifier string
-		err        error
-	)
-
-	for retry < maxRetry {
-		identifier, err = f()
-		if err == nil {
-			break
-		}
-		retry++
-	}
-	return identifier, err
-}
-
 // EC2Generator meta generator for EC2
 type EC2Generator struct {
 	baseURL *url.URL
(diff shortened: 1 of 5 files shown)


spec/cloud.go
```diff
@@ -58,17 +61,42 @@ func SuggestCloudGenerator(conf *config.Config) *CloudGenerator {
...
+		wg.Done()
+	}()
+	go func() {
+		wg.Wait()
+		close(gCh)
+	}()
+
+	var cg *CloudGenerator
+	for cg = range gCh {
+		break
```
itchyny: Need to call cancel() here?

spec/cloud.go
```diff
@@ -58,17 +61,42 @@ func SuggestCloudGenerator(conf *config.Config) *CloudGenerator {
...
+		wg.Done()
+	}()
+	go func() {
+		if isAzure(ctx) {
+			gCh <- &CloudGenerator{&AzureVMGenerator{azureVMBaseURL}}
+		}
+		wg.Done()
+	}()
+	go func() {
+		wg.Wait()
```
itchyny: I think wait group here does not work as expected, the first resolved CloudGenerator is returned before all `wg.Done()` is called.

itchyny:
I sent a pull request to the retry package to add retry with a context and now it's accepted. https://github.com/Songmu/retry/pull/3 Using this, we can cancel the retry sleep.

spec/cloud.go
astj: I think we don't have to wait all goroutines finish, because we only need first (and in most cases only) resolved CloudGenerator. So In original p-r I didn't wait `wg` synchronously.

But for simplification and with `retry.WithContext` (thank you!), I changed to wait synchronously as you suggest.

astj: ☝️ I tried in https://github.com/mackerelio/mackerel-agent/pull/516/commits/25dda67882e1a6eded26c063fa699a0eab15a38e but it caused data race, so I reverted back to asynchronous `wg.Wait()` at https://github.com/mackerelio/mackerel-agent/pull/516/commits/a6e43ab12663213cbe88f3063857a34d352e327f

itchyny:
LGTM!

spec/isec2_linux.go
```diff
@@ -35,11 +41,20 @@ func isEC2() bool {
 		return false
 	}
 
+	// give up if ctx already closed
+	if ctx.Err() != nil {
+		return false
+	}
```
itchyny:
```go
select {
  case <- ctx.Done():
  return false
  default:
} 
```

astj:
Thank you!

PR #519, opened 2018-08-30
add loadavg1 and loadavg15

itchyny:
This pull request adds loadavg1 and loadavg15 metrics to the loadavg graph, on all platforms except for Windows. The loadavg graph is more useful when comparing the loadavg5 vs 15 to see the CPU load is increasing or decreasing. We already updated the application to accept these system metrics.


itchyny:
Thank you.

PR #521, opened 2018-09-04
mackerel-agent.conf default permissions are too open

sudofox:
_Please note that I am not as familiar with packaging systems as you are, so the changes I have made are only suggestions and I have only tested them running on my own server (seem to work fine for me)._

The issue is that /etc/mackerel-agent/mackerel-agent.conf by default is too permissive, as the default permissions of 644 make it world-readable.

For example, here is the apache user on my machine, reading my Mackerel API key and my MySQL credentials:

```
[apache@node1 ~]# egrep "^[^#;](.*)(key|pass)" /etc/mackerel-agent/mackerel-agent.conf 
apikey = "censored"
command = "mackerel-plugin-mysql -username=root -password=censored"
```

The mackerel-agent service run under systemd is run as root, and the ownership of the file is also root. As such, I suggest that mackerel-agent should update the configuration file permissions to 600, and new installations should also have the permissions set to 600, to prevent credential leakage (especially in the case of a compromised box, in order to prevent privilege escalation.)

### DIFF
--- a/do_init.go
+++ b/do_init.go
@@ -39,7 +39,7 @@ func doInitialize(fs *flag.FlagSet, argv []string) error {
 		}
 		contents = append(contents, cBytes...)
 	}
-	return ioutil.WriteFile(*conffile, contents, 0644)
+	return ioutil.WriteFile(*conffile, contents, 0600)
 }
 
 type apikeyAlreadySetError string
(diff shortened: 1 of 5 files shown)


Songmu:
I am pleased with this pull request.

Your suggestion is completely correct, but unfortunately, we can not accept it.

Because it becomes incompatible change. For example, there are a certain number of Mackerel users using `mkr` command which reads `mackerel-agent.conf` with general user privileges.

If some users are concern about this permission, they could do the following correspondence.

- changing the permission by themselves to 600.
- by using `include` option, isolating the file in which the MySQL password is written.

sudofox:
Thank you for the detailed explanation! Now I understand the design of the default permissions. I will implement the include directive in my own system per your suggestion.

PR #526, opened 2018-10-07
To work in BusyBox

Songmu:
BusyBox's command system is slightly different from the general Linux environment. I support them.

I confirmed it works on my QNAP environment. Even though it may not be work in all BusyBox environments, we are not asking for it. I think that there is no problem because there is no incompatible change.

It does not mean the Mackerel.io officially support BusyBox as a business.


spec/linux/kernel.go
```diff
@@ -34,9 +33,15 @@ func (g *KernelGenerator) Generate() (interface{}, error) {
...
-		str := strings.TrimSpace(string(out))
-
-		results[key] = str
+		results[key] = strings.TrimSpace(string(out))
+	}
+	out, err := exec.Command("uname", "-o").Output()
+	if err == nil {
+		results["os"] = strings.TrimSpace(string(out))
+	} else {
+		// In some environments like busybox there is no `uname -o` option,
```
astj: [nit]
Recent busybox itself implements `uname -o`, therefore this comment should be "*old* busybox"

astj:
Please check or retry AppVeyor CI.

util/filesystem.go
```diff
@@ -43,20 +43,38 @@ var dfHeaderPattern = regexp.MustCompile(
 
 var logger = logging.GetLogger("util.filesystem")
 
+var dfOpt = "-Pkl"
+
+func init() {
+	// Some `df` command such as busybox does not have `-P` option.
```
astj: We should mention absence of `-l` too.
(Maybe busybox does not implement `-l` at all?)

util/filesystem.go
```diff
@@ -43,20 +43,38 @@ var dfHeaderPattern = regexp.MustCompile(
...
+	tio := &timeout.Timeout{
+		Cmd:       exec.Command("df", dfOpt),
+		Duration:  3 * time.Second,
+		KillAfter: 1 * time.Second,
+	}
+	exitSt, _, stderr, err := tio.Run()
+	if err == nil && exitSt.Code != 0 && strings.Contains(stderr, "df: invalid option -- P") {
```
astj: Maybe this will be broken when `dfOpt` is reordered like `-lPk` and/or error message of busybox changes, and I'm afraid that we'll miss the breakage because we don't have CI for busybox

util/filesystem.go
Songmu: Yes. The `df` command on my busybox environment also does not implement the `-l` option.

util/filesystem.go
Songmu: I do not plan to deal with busybox very seriously, so we do not have to worry about that. For now, I will remove the "P" here.

PR #529, opened 2018-10-10
[nit] Refactoring: use CheckConfig of mackerel-client-go

itchyny:
This is a part of refactoring towards removing the mackerel directory in favor of mackerel-client-go.
This pull request introduces CheckConfig of mackerel-client-go in the HostSpec struct. This struct itself will be replaced with the CreateHostParam struct of mackerel-client-go in the future.
Related: #515


astj:
lgtm

PR #531, opened 2018-10-23
Comment in source code updated

soudai:
I modified it because it is different from the actual behavior.

https://github.com/mackerelio/go-osstat/blob/master/memory/memory_linux.go#L70

If you have better comments please close this pull request and fix it there

### DIFF
--- a/metrics/linux/memory.go
+++ b/metrics/linux/memory.go
@@ -15,7 +15,8 @@ MemoryGenerator collect memory usage
 
 metric = "total", "free", "buffers", "cached", "active", "inactive", "swap_cached", "swap_total", "swap_free"
 
-Metrics "used" is calculated here like (total - free - buffers - cached) for ease.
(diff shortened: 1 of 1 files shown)


metrics/linux/memory.go
```diff
@@ -15,7 +15,8 @@ MemoryGenerator collect memory usage
 
 metric = "total", "free", "buffers", "cached", "active", "inactive", "swap_cached", "swap_total", "swap_free"
 
-Metrics "used" is calculated here like (total - free - buffers - cached) for ease.
+Metrics "used" is calculated here like (total - free - buffers - cached) or (total - available) or for ease.
+(total - available) is chosen when `MemAvailableEnabled` is enabled.
```
itchyny: Prefer `when MemAvailable is available in /proc/meminfo` because MemAvailableEnabled is automatically judged in the go-osstat library.

PR #532, opened 2018-10-24
CollectDfValues only from local file systems on linux

Songmu:
For example, when we are using NFS, the df command may hang and even
doesn't accept the "kill -kill" command, so I have targeted only the
local file system.

### DIFF
--- a/util/filesystem.go
+++ b/util/filesystem.go
@@ -54,7 +54,7 @@ func init() {
 	case "netbsd":
 		dfOpt = []string{"-Pkl"}
 	default:
-		dfOpt = []string{"-P"}
+		dfOpt = []string{"-Pl"}
 	}
 }
 


astj:
Yes this is incompatible that mackerel-agent will stop collecting non-local filesystem *specs*, but mackerel-agent has been filtering out non-local (not beginning with `/dev/`) filesystem *metrics* already, so imo this p-r does not introduce a big change.
https://github.com/mackerelio/mackerel-agent/blob/99e4e05b150cfd38f57a7790d608c10390cc4aff/metrics/filesystem.go#L30-L40

(And, in specs, mackerel host API returns filesystem specs as agent sends, but in web-view only local filesystems are displayed as "Filesystem" specs in right column, so even this incompatible change does not affect most users.)

image

PR #535, opened 2018-11-21
Fix disk metrics in linux 4.18+

itchyny:
4 more columns are added in /proc/diskstats in Linux 4.18+ so the agent fails to parse the diskstats due to the condition `len(values) != len(diskMetricsNames)`. See https://github.com/torvalds/linux/blob/v4.19/Documentation/ABI/testing/procfs-diskstats.


itchyny:
Thank you!

PR #539, opened 2018-12-13
Fix decoding error message of executables on Windows

mattn:
Plugins executable write UTF-8 bytes. But when executable does not exist, error message are encoded with the current codepage of command prompt. In japanese, it's cp932.

```
C:\>cmd /c deadbeef 2>&1 | nkf -g
Shift_JIS
```

This change add trick to decode output bytes to UTF-8. Run command with using "cmd /U" on Windows, and decode UTF-16 to UTF-8. If the output does not contains 0x00, it should be UTF-8 already.


mattn:
I changed timeout value to 300ms since I could not run tests on my PC.

itchyny:
LGTM! Thank you.

PR #540, opened 2018-12-17
Fix detecting EC2 instance on Windows

mattn:
Current implementation of Windows is trying to request ami-id endpoint even though the environment is not an EC2. It can do pre-checking before them as same as implementation of Linux.

https://docs.aws.amazon.com/ja_jp/AWSEC2/latest/WindowsGuide/identify_ec2_instances.html

### DIFF
--- a/spec/isec2_windows.go
+++ b/spec/isec2_windows.go
@@ -96,7 +96,7 @@ func isEC2UUID(uuid string) bool {
 		return true
 	}
 
-	// Check as littele endian.
+	// Check as little endian.
 	// see. https://docs.aws.amazon.com/ja_jp/AWSEC2/latest/UserGuide/identify_ec2_instances.html
 	fields := strings.Split(uuid, "-")
 	decoded, _ := hex.DecodeString(fields[0]) // fields[0]: UUID time_low(uint32)
--- a/spec/isec2_windows.go
+++ b/spec/isec2_windows.go
@@ -14,8 +14,8 @@ import (
 	"github.com/StackExchange/wmi"
 )
 
-// Win32_ComputerSystemProduct is struct for WMI. SKUNumber is nil-able.
-type Win32_ComputerSystemProduct struct {
+// Win32ComputerSystemProduct is struct for WMI. SKUNumber is nil-able.
+type Win32ComputerSystemProduct struct {
 	Caption           string
 	Description       string
 	IdentifyingNumber string
@@ -29,7 +29,7 @@ type Win32_ComputerSystemProduct struct {
 // If the OS is Windows, check UUID in WMI class Win32_ComputerSystemProduct first. If UUID seems to be EC2-ish, call the metadata API (up to 3 times).
 // ref. https://docs.aws.amazon.com/ja_jp/AWSEC2/latest/WindowsGuide/identify_ec2_instances.html
 func isEC2(ctx context.Context) bool {
-	var records []Win32_ComputerSystemProduct
+	var records []Win32ComputerSystemProduct
 	err := wmi.Query("SELECT * FROM Win32_ComputerSystemProduct", &records)
 	if err != nil {
 		return false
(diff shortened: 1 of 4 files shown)


spec/isec2_windows.go
```diff
@@ -0,0 +1,108 @@
...
+	Version           string
+}
+
+// If the OS is Windows, check UUID in WMI class Win32_ComputerSystemProduct first. If UUID seems to be EC2-ish, call the metadata API (up to 3 times).
+// ref. https://docs.aws.amazon.com/ja_jp/AWSEC2/latest/WindowsGuide/identify_ec2_instances.html
+func isEC2(ctx context.Context) bool {
+	var records []Win32ComputerSystemProduct
+	err := wmi.Query("SELECT * FROM Win32_ComputerSystemProduct", &records)
+	if err != nil {
+		return false
```
itchyny: Are you sure wmi query never fails on EC2 instance? If no it's better to fallback to the original check.

spec/isec2_windows.go
```diff
@@ -0,0 +1,108 @@
...
+	UUID              string
+	Vendor            string
+	Version           string
+}
+
+// If the OS is Windows, check UUID in WMI class Win32_ComputerSystemProduct first. If UUID seems to be EC2-ish, call the metadata API (up to 3 times).
+// ref. https://docs.aws.amazon.com/ja_jp/AWSEC2/latest/WindowsGuide/identify_ec2_instances.html
+func isEC2(ctx context.Context) bool {
+	var records []Win32ComputerSystemProduct
+	err := wmi.Query("SELECT * FROM Win32_ComputerSystemProduct", &records)
```
itchyny: Is collecting one column better than all columns in performance? If yes I think it's worth collecting only UUID column.

spec/isec2_windows.go
itchyny: (Well, not sure there's false negative cases due to wmi failure here)

spec/isec2_windows_test.go
```diff
@@ -0,0 +1,115 @@
...
+		{"45e12aec-dcd1-b213-94ed-01234abcdef", true}, // litte endian
+		{"45E12AEC-DCD1-B213-94ED-01234ABCDEF", true}, // litte endian
+		{"abcd1916-9099-7caf-fd21-01234abcdef", false},
+		{"ABCD1916-9099-7CAF-FD21-01234ABCDEF", false},
+		{"", false},
+	}
+
+	for _, tc := range tests {
+		if isEC2UUID(tc.uuid) != tc.expect {
+			t.Errorf("isEC2() should be %t: %q\n", tc.expect, tc.uuid)
```
itchyny: The format`%t` is weird but is the same mistake as isec2_linux_test.go so I don't care in this pull request.

itchyny: The format `%t` here looks weird but is the same mistake as isec2_linux_test.go so I don't care in this pull request.

spec/isec2_windows.go
mattn: Yes, I checked Win32_ComputerSystemProduct doesn't fail. Also this does not require any permisions to get. So I think the records of Win32_ComputerSystemProduct should be treated as same as files on Linux.

### isec2_linux.go
No|Condition|Check ami-id
-|-|-
1|Both files not exists. | No
2|Both files does not contains UUID starts with EC2. | No
3|A part of files contain UUID starts wtih EC2. | Yes

### isec2_windows.go
No|Condition|Check ami-id
-|-|-
1|Both records can not be given | No (as same as 1 above)
2|Both records does not contains UUID starts with EC2 | No (as same as 2 above)
3|A part of records contain UUID starts wtih EC2 | Yes

spec/isec2_windows.go
mattn: Okay, will fix.

spec/isec2_windows.go
itchyny: Confirmed, thanks.

itchyny:
Thank you!

PR #544, opened 2019-01-30
fix counter naming problem on Windows

lufia:
Some NICs contains **'/'** in model name. **'/'** should replace with **'_'**.


itchyny:
LGTM

PR #549, opened 2019-03-26
support Go Modules

lufia:
I made to support Go Modules.


itchyny:
appveyor.yml is not still updated?

lufia:
I updated appveyor.yml for using both Go 1.12 and Go Modules.

PR #551, opened 2019-03-28
Reduce fetch ec2 metadata count

i2tsuki:
There is an EC2 Instance Identity Document Endpoint in EC2 Metadata Endpoint to gets instance metadata.

Getting instance metadata from `169.254.169.254/latest/dynamic/instance-identity/document`, we can get many metadata at once. Also, although not implemented in this PR, metadata can be validated using signature, so that if implemented in the future, it can be validated.

### DIFF
--- a/spec/cloud.go
+++ b/spec/cloud.go
@@ -165,25 +165,66 @@ type EC2Generator struct {
 	baseURL *url.URL
 }
 
+// EC2Instance struct for EC2 instance identity document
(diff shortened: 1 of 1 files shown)


itchyny:
This is a good idea, if you have some trouble to fix we'll take over, thanks.

itchyny:
Closing due to inactivity.

PR #553, opened 2019-03-29
migrate to use mkr.Interface instead of mackerel.NetInterface

lufia:
This p-r is a part of an integration mackerel.* into mackerel-client-go.

### DIFF
--- a/spec/freebsd/interface.go
+++ b/spec/freebsd/interface.go
@@ -3,7 +3,6 @@
 package freebsd
 
 import (
-	"github.com/mackerelio/mackerel-agent/spec"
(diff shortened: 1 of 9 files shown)


spec/freebsd/interface.go
```diff
@@ -2,14 +2,16 @@
 
 package freebsd
 
-import "github.com/mackerelio/mackerel-agent/spec"
+import (
+	mkr "github.com/mackerelio/mackerel-client-go"
```
itchyny: We would like to use mackerel alias, because mackerel-client-go is not mkr. It's ok until it conflicts the mackerel package inside this repository.

PR #562, opened 2019-05-11
[nit] Update help URL in mackerel-agent.sample.conf

astj:
Although we can still access to old url with redirect, but we should use preferred url.

### DIFF
--- a/mackerel-agent.sample.conf
+++ b/mackerel-agent.sample.conf
@@ -11,7 +11,7 @@
 # ignore = "/dev/ram.*"
 
 # Configuration for Custom Metrics Plugins
-# see also: http://help-ja.mackerel.io/entry/advanced/custom-metrics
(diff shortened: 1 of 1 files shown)


mackerel-agent.sample.conf
```diff
@@ -11,7 +11,7 @@
 # ignore = "/dev/ram.*"
 
 # Configuration for Custom Metrics Plugins
-# see also: http://help-ja.mackerel.io/entry/advanced/custom-metrics
+# see also: https://mackerel.io/ja/docs/entry/advanced/custom-metrics
```
susisu: Is it ok that the Japanese document is referenced here?

susisu: Also here...
https://github.com/mackerelio/mackerel-agent/blob/d9e3082a32b96c17560a375e5e78babcb0f34e8d/wix/mackerel-agent.sample.conf#L9-L10
Anyway, they can be fixed in another pr.

PR #563, opened 2019-05-13
migrate to use mkr.CheckReports instead of monitoringChecksPayload

lufia:
I replaced mackerel.monitoringChecksPayload with mkr.CheckReports.

### DIFF
--- a/mackerel/check.go
+++ b/mackerel/check.go
@@ -1,8 +1,6 @@
 package mackerel
 
 import (
-	"encoding/json"
-
 	"github.com/mackerelio/mackerel-agent/checks"
 	mkr "github.com/mackerelio/mackerel-client-go"
 )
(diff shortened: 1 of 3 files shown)


mackerel/check.go
```diff
@@ -44,17 +19,31 @@ func (api *API) ReportCheckMonitors(hostID string, reports []*checks.Report) err
...
 			Name:                 report.Name,
-			Status:               report.Status,
+			Status:               mkr.CheckStatus(report.Status),
 			Message:              msg,
-			OccurredAt:           Time(report.OccurredAt),
-			NotificationInterval: report.NotificationInterval,
-			MaxCheckAttempts:     report.MaxCheckAttempts,
+			OccurredAt:           report.OccurredAt.Unix(),
+			NotificationInterval: normalize(report.NotificationInterval, NotificationIntervalFallback),
+			MaxCheckAttempts:     normalize(report.MaxCheckAttempts, 0),
```
susisu: The minimum value of `maxCheckAttempts` is `1`.
https://mackerel.io/api-docs/entry/check-monitoring#post
(`0` will be omitted from JSON output and the default value `1` will be used in such a case, so there are no problems, though)

mackerel/check.go
```diff
@@ -1,41 +1,16 @@
...
-}
-
-// MarshalJSON implements json.Marshaler.
-func (h monitorTargetHost) MarshalJSON() ([]byte, error) {
-	return json.Marshal(map[string]string{
-		"type":   "host",
-		"hostId": h.HostID,
-	})
-}
+const NotificationIntervalFallback = 10
```
susisu: `make lint` fails. Please add some comments.

lufia: sorry, I resolved it.

lufia:
Thank you.

PR #565, opened 2019-05-15
migrate CreatingMetricsValue to mackerel.HostMetricValue

lufia:
I replaced CreatingMetricsValue with HostMetricValue of mackerel-client.

And also I dropped **"s"** from *PostMetric **"s"** Values* because type name don't contain **"s"**.


lufia:
Thank you!

PR #571, opened 2019-06-26
[check-plugin] Support custom_identifier 

astj:
configuration syntax is same as metrics plugin:
```toml
[plugin.checks.foobar]
custom_identifier = "myapp.example.com"
command = "check-uptime"
```

### DIFF
--- a/command/command.go
+++ b/command/command.go
@@ -518,23 +518,45 @@ func runCheckersLoop(app *App, termCheckerCh <-chan struct{}, quit <-chan struct
 			logger.Debugf("RunChekcerLoop: Extend the delay to %d seconds. There are %d reports.", reportCheckDelay, len(reports))
 		}
 
-		partialReports := make([]*checks.Report, 0, checkReportMaxSize)
-		for i, report := range reports {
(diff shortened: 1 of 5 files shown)


command/command.go
```diff
@@ -518,23 +518,38 @@ func runCheckersLoop(app *App, termCheckerCh <-chan struct{}, quit <-chan struct
...
 	}
 }
 
-func reportCheckMonitors(app *App, reports []*checks.Report) {
+func reportCheckMonitors(app *App, customIdentifier string, reports []*checks.Report) {
+	host := app.Host
+	if customIdentifier != "" {
+		if h, ok := app.CustomIdentifierHosts[customIdentifier]; ok {
+			host = h
+		}
```
itchyny: When host of custom identifier is not found, the report is posted to app.Host but it is inconsistent with metric plugins. See https://github.com/mackerelio/mackerel-agent/blob/5d4632d/command/command.go#L379-L385. Is it ok?

astj: Not ok!!!!!!! Thank you for review.
The behavior should be same as metric plugins.

PR #575, opened 2019-07-17
Generate and include CREDITS file in the release artifacts

itchyny:
We should clarify the licenses of the library we use. In this pull request, we generate CREDITS file on crossbuild to ensure the file collected by [goxz](https://github.com/Songmu/goxz).


itchyny:
Thank you!

PR #576, opened 2019-07-17
[nit] add update target in Makefile

itchyny:
So that we can update the libraries and don't forget to keep tidy.


itchyny:
Thanks.

PR #557, opened 2019-04-21
Add mips and arm64 architecture debian packaging support

tnishinaga:
Hello,

I add mips and arm64 architecture support to mackerel debian packaging.

I tested on EdgeRouterLite(mips) and SynQuacer(arm64), it works fine 😄 .

PR #577, opened 2019-07-18
Fix to fit #557 into our workflow.

hayajo:
Ref #557.
Thank you, @tnishinaga

- Build arm64 and mips binary at the crossbuild-package task
- Build deb packeages for arm64 and mips at the dev-v2 task
- Build deb packages in Dockerlized environment


packaging/deb-systemd/debian/control
```diff
@@ -9,7 +9,7 @@ Vcs-Git: https://github.com/mackerelio/mackerel-agent.git
 Homepage: https://mackerel.io
 
 Package: mackerel-agent
-Architecture: amd64
+Architecture: any
```
astj: ref: https://www.debian.org/doc/debian-policy/ch-controlfields.html#architecture

When `any`, produced binary package will have specific architecture (amd64, mips, etc…), which is desired behavior.

astj:
@hayajo (and @tnishinaga ) This p-r and #557 will build only systemd-based deb package for arm and mips. I hope this will satisfy major use cases, but in some cases it might not fit. For example, if you want to install mackerel-agent into some limited environment which does not use systemd, like NAS/router with busybox, this deb will not work since they works on top of systemd.
Do you think current artifacts are enough for use cases?

.gitignore
```diff
@@ -18,5 +18,10 @@ packaging/deb/debian/mackerel-agent.bin
 packaging/*.tar.gz
 packaging/deb-build
 packaging/rpm-build
+packaging/*.deb
```
astj: Hmm, should we change directory structure to be more gitignore-friendly?

.travis.yml
```diff
@@ -9,13 +9,6 @@ env:
   - PATH=/home/travis/gopath/bin:$PATH DEBIAN_FRONTEND=noninteractive
   - secure: "VG3xbEjFSuB80QBFchHZqQDB4VF74RTXaNSBPl/bavGr9My2RSqtfXqfEGUHFwo5zPM7LA0bIEC7jsobf6M6vtfVgJKOOXEfF/pfD83n++eHZe0yLN6Wysp1Bn14KH+ont550gDityORqEvX3pK4k9hSb8dXpjGnCtrAwqAS+eg="
   - GO111MODULE=on
-addons:
```
astj: 😄

.gitignore
hayajo: It will be nice to put the artifacts in one directory.
But we also need to fix other tools, so let's discuss in another issue.

PR #578, opened 2019-07-18
Change deb package maintainer name to mackerel

astj:
Now maintainer name is "Hatena" in debian control file, but in changelog author name is "mackerel".
https://github.com/mackerelio/mackerel-agent/blob/master/packaging/deb/debian/changelog#L16

Therefore currently [changelog-should-mention-nmu](https://lintian.debian.org/tags/changelog-should-mention-nmu.html) warning is claimed by lintian.
This p-r changes maintainer name to match those two.  (I believe changing maintainer name will not cause problem, but I need further investigation.)

### DIFF
--- a/packaging/deb-systemd/debian/control
+++ b/packaging/deb-systemd/debian/control
@@ -1,7 +1,7 @@
 Source: mackerel-agent
 Section: admin
 Priority: extra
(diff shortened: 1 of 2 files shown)


astj:
We should consider this issue in future, but not now.

PR #580, opened 2019-07-23
avoid to use unnamed NICs for registering hosts on Windows

lufia:
*Registering host information* API will reject registering a host that have an unnamed interface because *name* parameter is not optional.

https://mackerel.io/api-docs/entry/hosts

Thus I changed to skip unnamed NICs.

I'm not sure why *ai.Description* or *ifi.Name* is set as empty string by Windows.

### DIFF
--- a/metrics/windows/interface.go
+++ b/metrics/windows/interface.go
@@ -62,7 +62,7 @@ func NewInterfaceGenerator(interval time.Duration) (*InterfaceGenerator, error) 
 	names := []string{}
 	for ai = first; ai != nil; ai = ai.Next {
 		name, err := windows.AnsiBytePtrToString(&ai.Description[0])
-		if err == nil {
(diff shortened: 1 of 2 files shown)


metrics/windows/interface.go
```diff
@@ -62,7 +62,7 @@ func NewInterfaceGenerator(interval time.Duration) (*InterfaceGenerator, error)
 	names := []string{}
 	for ai = first; ai != nil; ai = ai.Next {
 		name, err := windows.AnsiBytePtrToString(&ai.Description[0])
-		if err == nil {
+		if err == nil && name != "" {
```
itchyny: This code is used for collecting metrics, not used when registering host information, right?

lufia: I added same code to `Generate` in spec. Thank you.

PR #583, opened 2019-07-23
Check command name on pid check for pid confliction after OS restart

itchyny:
On unexpected OS restart, there happens that different process uses the same pid as mackerel-agent before restart. We should start the mackerel-agent, when it finds a process matching the old pid file but is not mackerel-agent.


itchyny:
Thank you.

PR #584, opened 2019-07-24
Remove memory.active and inactive metrics

itchyny:
These metrics is not used in Mackerel web console and users suffer a disadvantage in a viewpoint of metrics count.


itchyny:
Thank you.

PR #585, opened 2019-07-24
Allow working directory configuration in env of metadata plugins

itchyny:
The `MACKEREL_PLUGIN_WORKDIR` environment variable is handled by the plugin process for metrics and check plugins. So this value is often set in the `env` config within mackerel-agent.conf.
However, this environment variable is handled by the mackerel-agent process for metadata plugin. This is not intuitive when user want to configure the working directory for each metadata plugins.

### DIFF
--- a/command/metadata.go
+++ b/command/metadata.go
@@ -17,14 +17,10 @@ func metadataGenerators(conf *config.Config) []*metadata.Generator {
 
 	workdir := filepath.Join(pluginutil.PluginWorkDir(), "mackerel-metadata")
 	for name, pluginConfig := range conf.MetadataPlugins {
-		workdir := workdir
-		if dir := lookupPluginWorkDir(pluginConfig.Command.Env); dir != "" {
-			workdir = dir
-		}
 		generator := &metadata.Generator{
 			Name:      name,
 			Config:    pluginConfig,
-			Cachefile: filepath.Join(workdir, name),
+			Cachefile: getCacheFileName(name, workdir, pluginConfig),
 		}
 		logger.Debugf("Metadata plugin generator created: %#v %#v", generator, generator.Config)
 		generators = append(generators, generator)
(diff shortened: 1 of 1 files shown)


command/metadata.go
```diff
@@ -14,12 +15,16 @@ import (
 func metadataGenerators(conf *config.Config) []*metadata.Generator {
 	generators := make([]*metadata.Generator, 0, len(conf.MetadataPlugins))
 
-	workdir := pluginutil.PluginWorkDir()
+	workdir := filepath.Join(pluginutil.PluginWorkDir(), "mackerel-metadata")
```
susisu: [note] `pluginutil.PluginWorkDir()` does look `MACKEREL_PLUGIN_WORKDIR`. 
https://github.com/mackerelio/golib/blob/master/pluginutil/tempfile.go

command/metadata.go
```diff
@@ -14,12 +15,16 @@ import (
...
 	for name, pluginConfig := range conf.MetadataPlugins {
+		workdir := workdir
+		if dir := lookupPluginWorkDir(pluginConfig.Command.Env); dir != "" {
+			workdir = dir
+		}
 		generator := &metadata.Generator{
 			Name:      name,
 			Config:    pluginConfig,
-			Cachefile: filepath.Join(workdir, "mackerel-metadata", name),
+			Cachefile: filepath.Join(workdir, name),
```
susisu: Is it ok to not append `mackerel-metadata/` to the path when `MACKEREL_PLUGIN_WORKDIR` is configured per plugin?
`filepath.Join(workdir, name)` may conflict with the plugin executable?

itchyny: The temporary file of metric plugin is stored at `MACKEREL_PLUGIN_WORKDIR/pluginname` (https://github.com/mackerelio/go-mackerel-plugin/blob/7d51210247a3f1a73fdd43098cda470b328cb58f/mackerel-plugin.go#L177) so I think this is consistent. The value is not expected to be set to the bin directory.

susisu: I understand!

itchyny: I reconsidered this and prepended `mackerel-plugin-metadata-` to the plugin name. The file without the env variable config is not changed.

itchyny:
Thank you.

PR #586, opened 2019-07-24
[nit] Introduce context.Context

itchyny:
This is a part of refactoring, introducing context.Context. Actually mackerel-agent was written before context package came out. I believe this is useful for canceling, especially for testing, as proved in mackerel-container-agent.


itchyny:
Thank you.

PR #587, opened 2019-07-25
change the owner of files created in docker

hayajo:
change the owner to operate the files that are created in the docker.


hayajo:
Thank you!

PR #592, opened 2019-09-09
Fixed to create configuration directory directory when executing init command if not exist directory

homedm:
If the configuration file directory does not exist, 
executing ```mackerel-agent init -apikey "api key "``` will return an error with no such file or directory.
So, modified to create a directory if it doesn't exist when executing the above command,
And, add test code for not existing configuration directory.

### DIFF
--- a/do_init.go
+++ b/do_init.go
@@ -28,7 +28,7 @@ func doInitialize(fs *flag.FlagSet, argv []string) error {
 		if !os.IsNotExist(err) {
 			return err
 		}
-		err := os.MkdirAll(filepath.Dir(*conffile), 0744)
+		err := os.MkdirAll(root, 0744)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


do_init.go
```diff
@@ -20,6 +21,19 @@ func doInitialize(fs *flag.FlagSet, argv []string) error {
...
 		return fmt.Errorf("-apikey option is required")
 	}
+
+	// make the config file directory if not exist
+	root := filepath.Dir(*conffile)
+	if _, err := os.Stat(root); err != nil {
+		if !os.IsNotExist(err) {
+			return err
+		}
+		err := os.MkdirAll(root, 0744)
```
astj: Is there any reason to choose 0744, not 0755?
Currently we use 0755 for newly created directories (explicitly in some cases, or implicitly in many cases). So I  prefer 0755 here if there's no special reason.

homedm: OK. thank you for reviewing

PR #599, opened 2019-10-21
Add armhf Debian package to release

hnw:
Add armhf(32bit ARM with hardfloat) Debian package support.

The package has already been tested in two weeks on my own Raspberry Pi 2, and it works fine.

However, this pull request has one problem. This change requires Docker image `yhnw/docker-mackerel-deb-builder`. We should use Docker image which is managed by Mackerel dev team.

If the dev team approve mackerelio/docker-mackerel-deb-builder#1, then I'll fix the problem.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,7 +94,7 @@ cover: deps
 # We should be installed tools of native architecture.
 .PHONY: crossbuild-package
 crossbuild-package: deps
-	mkdir -p ./build-linux-386 ./build-linux-amd64 ./build-linux-arm64 ./build-linux-mips
+	mkdir -p ./build-linux-386 ./build-linux-amd64 ./build-linux-arm64 ./build-linux-mips ./build-linux-armhf
 	GOOS=linux GOARCH=386 make build
 	mv build/$(MACKEREL_AGENT_NAME) build-linux-386/
 	GOOS=linux GOARCH=amd64 make build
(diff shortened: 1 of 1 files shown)


astj:
[memo]
- In Debian "armhf" supports ARMv7~ https://www.debian.org/ports/arm/
  - "armhfp" for CentOS also supports ARMv7~ https://wiki.centos.org/SpecialInterestGroup/AltArch/armhfp
- From those point of view, ARMv7 seems enough, but:
- Raspberry Pi 1 (and zero?) are ARMv6 https://en.wikipedia.org/wiki/Raspberry_Pi#Specifications
- For those ARMv6 family, Raspberry Pi teams maintain Debian fork (Raspbian) for armhf with ARMv6 https://www.raspbian.org/RaspbianFAQ#What_is_Raspbian.3F

astj:
@hnw Thank you for pull request!  Adding 32bit arm packages seems good, but maybe we need some discussion around GOARM parameter. Please check the review comment above.

hnw:
Hi @astj. Thanks for the comment.

I fixed a Docker image and added comments.

As you mentioned above, ARMv6 is called as armhf on Raspbian. Raspberry Pi is widely used, so I think we should support them.

astj:
Sorry to have kept you waiting so long. LGTM!

PR #609, opened 2019-11-25
refactor: improve interface and testing for spec/cloud

astj:
- Ensure `spec.CloudMetaGenerator#Generate()` generates `*mackerel.Cloud`
  - Currently (fortunately) it does
  - The return value will be wrapped in `spec.CloudGenerator#Generate()` as `interface{}`, in order to satisfy `spec.Generator#Generate`
- Test`spec.CloudGenerator#Generate()` and `spec.EC2Generator#Generate()` separetely
  - Now the former uses `mockCloudMetaGenerator`

### DIFF
--- a/spec/cloud_test.go
+++ b/spec/cloud_test.go
@@ -114,7 +114,7 @@ func TestEC2Generate(t *testing.T) {
 	if err != nil {
 		t.Errorf("should not raise error: %s", err)
 	}
-	g := &CloudGenerator{&EC2Generator{u}}
+	g := &EC2Generator{u}
 
 	value, err := g.Generate()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


spec/cloud_test.go
```diff
@@ -54,8 +57,59 @@ func (g *mockGCECloudMetaGenerator) IsGCE(ctx context.Context) bool {
...
+	if err != nil {
+		t.Errorf("should not raise error: %s", err)
+	}
+
+	cloud, typeOk := value.(*mackerel.Cloud)
+	if !typeOk {
+		t.Errorf("value should be *mackerel.Cloud. %+v", value)
+	}
+
+	if cloud.Provider != "mock" {
```
lufia: [nit] this expression will panic if previous test was error. but i think ok because it may not be happen.

astj: Should we use `t.Fatalf` instead?

lufia: IMHO, no. `t.Fatalf` can use only when the prerequisites are failed.

I suggest to use `return` statement or `if cloud != nil && cloud.Provider...` if you fix politely.

astj: Okay. I'll fix TestCloudGenerate and TestEC2Generator (, but keep others as is).

astj:
thank you for reviewing!

PR #618, opened 2019-12-09
Implement GCEGenerator.SuggestCustomIdentifier

tanatana:
This P-R implements GCEGenerator.SuggestCustomIdentifier.


command/command_test.go
```diff
@@ -95,6 +95,9 @@ func TestPrepareWithCreate(t *testing.T) {
 	conf, mockHandlers, ts, deferFunc := newMockAPIServer(t)
 	defer deferFunc()
 
+	// To prevent unexpected behavior in the TravisCI environment
```
astj: If we don't have interests about cloud platforms (I suspect so, but not sure), can we set this in `newMockAPIServer`. How do you think?

command/command_unix_test.go
```diff
@@ -43,7 +43,7 @@ func TestRunOnce(t *testing.T) {
 func TestRunOncePayload(t *testing.T) {
 	if testing.Short() {
 		origMetricsInterval := metricsInterval
-		metricsInterval = 1
+		metricsInterval = 1 * time.Second
```
astj: 😱

command/command_test.go
tanatana: Well, I want to fix only the test cases that fall due to SSRF-like behavior, so I think it is better that inserted them in ad hoc.

spec/cloud.go
```diff
@@ -314,9 +314,41 @@ func (g gceMeta) toGeneratorResults() *mackerel.Cloud {
...
-// SuggestCustomIdentifier for GCE is not implemented yet
+// SuggestCustomIdentifier the identifier of the GCE VM instance
 func (g *GCEGenerator) SuggestCustomIdentifier() (string, error) {
-	return "", nil
+	identifier := ""
+	err := retry.Retry(3, 2*time.Second, func() error {
+		cl := httpCli()
+		// $ curl -q http://metadata.google.internal/computeMetadata/v1/instance/id -H "Metadata-Flavor: Google"
+		// 8618997027285493035 # This is a unique, numerical ID that is generated by Compute Engine.
+		// https://cloud.google.com/compute/docs/storing-retrieving-metadata#default
```
lufia: good comment. I like this.

spec/cloud.go
```diff
@@ -122,7 +122,7 @@ var CloudGeneratorSuggester *cloudGeneratorSuggester
 
 func init() {
 	ec2BaseURL, _ = url.Parse("http://169.254.169.254/latest/meta-data")
-	gceMetaURL, _ = url.Parse("http://metadata.google.internal./computeMetadata/v1/?recursive=true")
```
astj: Without `/?recursive=true`, we cannot get GCE Metadata in `requestGCEMeta`.

```
astj@instance-1:~$ curl -H 'Metadata-Flavor: Google' http://metadata.google.internal./computeMetadata/v1
/computeMetadata/v1/
astj@instance-1:~$ curl -H 'Metadata-Flavor: Google' 'http://metadata.google.internal./computeMetadata/v1/?recursive=true'
{"instance":{"attributes":{"ssh-keys": ... (snip)
```

tanatana: Oops, It's my fault. I'm going add `?recursive=true` in requestGCEMeta.

spec/cloud.go
```diff
@@ -314,9 +314,41 @@ func (g gceMeta) toGeneratorResults() *mackerel.Cloud {
...
+// SuggestCustomIdentifier the identifier of the GCE VM instance
 func (g *GCEGenerator) SuggestCustomIdentifier() (string, error) {
-	return "", nil
+	identifier := ""
+	err := retry.Retry(3, 2*time.Second, func() error {
+		cl := httpCli()
+		// $ curl -q http://metadata.google.internal/computeMetadata/v1/instance/id -H "Metadata-Flavor: Google"
+		// 8618997027285493035 # This is a unique, numerical ID that is generated by Compute Engine.
+		// https://cloud.google.com/compute/docs/storing-retrieving-metadata#default
+		req, err := http.NewRequest("GET", g.metaURL.String()+"/instance/id", nil)
```
astj: I guess instance id is already retrieved in `GCEGenerator#Generate`, since `requestGCEMeta` retrieves GCE metadata recursively.
You can reduce http request to GCE metadata endpoint by sharing response between `#Generate` and `#SuggestCustomIdentifier`.

tanatana: As you said, the metadata retrieval logic is duplicated.
I changed the process to cache the value retrieved by `Generate ()`.

spec/cloud.go
```diff
@@ -314,9 +320,20 @@ func (g gceMeta) toGeneratorResults() *mackerel.Cloud {
...
 
-// SuggestCustomIdentifier for GCE is not implemented yet
+func builtGCECustomIdentifier(instanceID string) string {
+	return instanceID + ".gce.cloud.google.com"
+}
+
+// SuggestCustomIdentifier the identifier of the GCE VM instance
 func (g *GCEGenerator) SuggestCustomIdentifier() (string, error) {
-	return "", nil
+	if g.meta.Instance == nil || g.meta.Instance.InstanceID == 0 {
```
astj: Should we consider the case `g.meta == nil`?

tanatana: We don't have to check `g.meta == nil` case because g.meta is not reference type. So always g.meta have value even if Generate() was not called.

tanatana:
Thank you!

PR #624, opened 2020-01-22
Allow formatted duration in config

itchyny:
The unit of `notification_interval` is in minutes but it is hard to tell from the config (and we actually made mistake). How about allowing time.Duration here? For example,
```toml
[plugin.checks.sample]
command = "sample.sh"
notification_interval = "1h30m"
```
this is much readable :).

Here's a few notes.

- The formatted duration requires quotes. This is due to the TOML spec.
  - NG: `notification_interval: 1h`, OK: `notification_interval: "1h"`
- Units for day (`1d` or whatever) is not allowed. This is due to the Go language spec.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -127,9 +127,9 @@ type Config struct {
 // PluginConfig represents a plugin configuration.
 type PluginConfig struct {
 	CommandConfig
-	NotificationInterval  *int32        `toml:"notification_interval"`
-	CheckInterval         *int32        `toml:"check_interval"`
-	ExecutionInterval     *int32        `toml:"execution_interval"`
+	NotificationInterval  *duration     `toml:"notification_interval"`
+	CheckInterval         *duration     `toml:"check_interval"`
+	ExecutionInterval     *duration     `toml:"execution_interval"`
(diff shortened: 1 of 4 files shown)


config/config.go
```diff
@@ -289,8 +289,8 @@ func (pconf *PluginConfig) buildCheckPlugin(name string) (*CheckPlugin, error) {
 	plugin := CheckPlugin{
 		Command:               *cmd,
 		CustomIdentifier:      pconf.CustomIdentifier,
-		NotificationInterval:  pconf.NotificationInterval,
-		CheckInterval:         pconf.CheckInterval,
+		NotificationInterval:  pconf.NotificationInterval.toMinutes(),
+		CheckInterval:         pconf.CheckInterval.toMinutes(),
```
itchyny: I'm wondering we should change CheckInterval to time.Duration or not. But maybe we can do this in later pull request (if we want).

- Change `type duration int32` to `type duration time.Duration`
- Change `CheckInterval *int32` to `CheckInterval time.Duration`
- Make https://github.com/mackerelio/mackerel-agent/blob/fe40e3c5a506ce552657789e3a32259878aaf790/checks/checker.go#L92 simple.

PR #626, opened 2020-01-23
fix typo

shogo82148:


lufia:
@shogo82148 
Thank you.
I've added a commit to fix a other typo. I will merge this if tests are completed.

PR #635, opened 2020-03-01
fix too late closing the response body

shogo82148:
`resp.Body` should be closed before next request,
but actually they are closed after all requests sent.
because `defer resp.Body.Close()` will be invoked in the end of the function,
not the end of the scope.

```go
	for _, key := range metadataKeys {
		// snip
		resp, err := cl.Do(req)
		if err != nil {
			cloudLogger.Debugf("This host may not be running on EC2. Error while reading '%s'", key)
			return nil, nil
		}
		defer resp.Body.Close()
		// snip
		// resp.Body.Close() should be invoked HERE.
	}

	// actually they are invoked HERE.
	return &mackerel.Cloud{Provider: "ec2", MetaData: metadata}, nil
}
```

### DIFF
--- a/spec/cloud.go
+++ b/spec/cloud.go
@@ -259,34 +259,44 @@ func (g *EC2Generator) Generate() (*mackerel.Cloud, error) {
 	metadata := make(map[string]string)
 
 	for _, key := range metadataKeys {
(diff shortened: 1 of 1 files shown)


astj:
Oh, you're right. Thank you for nice pull request.

PR #640, opened 2020-04-26
Improve FreeBSD rc script

metalefty:
mackerel-agent is now available as [FreeBSD ports](https://www.freshports.org/sysutils/mackerel-agent/).  Here's a feedback on rc script from a FreeBSD developer.
- Add some comment
- Do not spit out logs on stdout but use syslog when run as daemon
- Use the same config file path as well as other distro (mackerel-agent subdir)
- Run as supervisor mode

### DIFF
--- a/packaging/freebsd/README.md
+++ b/packaging/freebsd/README.md
@@ -1,22 +1,67 @@
 mackerel-agent for FreeBSD
 ==========================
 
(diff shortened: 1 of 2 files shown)


metalefty:
CC: @owatan

metalefty:
@susisu I've totally rewritten the README.

PR #667, opened 2020-11-04
[wip] retry once on network error on requesting Mackerel API

astj:
https://github.com/mackerelio/mackerel-client-go/pull/112


astj:
=> #669

PR #669, opened 2020-11-09
Retry once immediately on posting metrics and check reports when error is caused by net/http (network error)

astj:
ref #667

### DIFF
--- a/command/command.go
+++ b/command/command.go
@@ -353,23 +353,16 @@ func loop(app *App, termCh chan struct{}) error {
 }
 
 func postHostMetricValuesWithRetry(app *App, postValues []*mkr.HostMetricValue) (err error) {
-	ctx, cancel := context.WithTimeout(context.TODO(), 25*time.Second)
-	defer cancel()
+	deadline := time.Now().Add(25 * time.Second)
 
 	err = app.API.PostHostMetricValues(postValues)
 	if err == nil {
 		logger.Debugf("Posting metrics succeeded.")
 		return
 	}
-	// Do not retry when first request already took long.
-	// (Typically when first request has timeout)
-	if ctxErr := ctx.Err(); ctxErr != nil {
-		logger.Warningf("Failed to post metrics value (will retry): %s", err.Error())
-		return
-	}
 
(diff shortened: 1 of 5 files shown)


command/command.go
```diff
@@ -552,24 +579,51 @@ func reportCheckMonitors(app *App, customIdentifier string, reports []*checks.Re
...
+
+	err = app.API.ReportCheckMonitors(hostID, reports)
+	if err == nil {
+		logger.Debugf("ReportCheckMonitors: succeeded")
+		return
+	}
+	// Do not retry when first request already took long.
+	// (Typically when first request has timeout)
+	if ctxErr := ctx.Err(); ctxErr != nil {
+		logger.Warningf("ReportCheckMonitors: failed to report (will retry): %s", err)
```
lufia, line 610: Should this use `ctxErr` instead of `err`?

command/command.go
```diff
@@ -354,6 +352,35 @@ func loop(app *App, termCh chan struct{}) error {
...
+	if err == nil {
+		logger.Debugf("Posting metrics succeeded.")
+		return
+	}
+	// Do not retry when first request already took long.
+	// (Typically when first request has timeout)
+	if ctxErr := ctx.Err(); ctxErr != nil {
+		logger.Warningf("Failed to post metrics value", err.Error())
+		return
+	}
```
lufia, line 369: I recommend to use always `err`.

command/command.go
astj: Here `ctx` is used just for handling timeout, so there's no useful information for logging.  So I'd like to log original err.
(So should I use just something like `start := time.Now(); ... if time.Now().After(start.Add25 * time.Second)) { ... }` to avoid confusion?)

astj: Okay, I rewrote not to use `ctx`

command/command.go
astj: Oh, this was just my mistake.  Fixed log messages.
(And our `golib/logging` https://pkg.go.dev/github.com/mackerelio/golib@v1.0.0/logging don't have `logging.Warning` 😄

lufia (approved):
Looks good.

PR #676, opened 2020-11-19
Fix artifact filename pattern again to include mackerel-agent_{os}_{arch}.tar.gz to GitHub Release artifacts

astj:
#674 did not solve the problem.  It seems we cannot use `{}` here


yseto (approved):
LGTM!

PR #684, opened 2020-12-21
Build Windows package on GitHub Actions

yseto:
I added GitHub Actions for CI(Windows), Build Package(Windows)


.github/workflows/go.yml
```diff
@@ -59,8 +106,140 @@ jobs:
...
+      run: git fetch --prune --unshallow --tags
+
+    - uses: actions/download-artifact@v2
+      with:
+        name: windows-build-artifacts-${{ matrix.PLATFORM_ID }}
+        path: build/
+
+    - name: Build Installer
+      run: |
+        $VERSION = git describe --tags --abbrev=0 --match "v*" | %{ $_ -replace "v","" }
```
astj, line 217: I think this does not work as expected.
- Currently, version tag was created via GitHub Release by [`mackerel-github-release` command](https://github.com/mackerelio/golib/blob/master/cmd/mackerel-github-release/main.go#L128), invoked in ["Release to GitHub Releases" workflow job](https://github.com/mackerelio/mackerel-agent/blob/master/.github/workflows/go.yml#L85)
- This job runs BEFORE "Release to GitHub Releases" job, so desired (new) version tag does not exist

Actually, current workflow runs as following (complicated :brain:)
- Version bump p-r merged (like #682, here we assume 0.70.3 => 0.71.0)
- On this merge commit at master branch:
  - GitHub Actions runs and
    - build Linux v0.71.0 artifacts
    - creates "v0.71.0" GitHub Releases and git tags
    - upload Linux artifacts to "v0.71.0" GitHub releases
  - AppVeyor tests and builds "v0.70.3", but does not upload releases
    - https://ci.appveyor.com/project/mackerelio/mackerel-agent/builds/36818983
    - At this timing, latest git tag is still "v0.70.3"
- On "v0.71.0" tags pushed:
  - GitHub Actions does nothing since conditions are not met https://github.com/mackerelio/mackerel-agent/blob/master/.github/workflows/go.yml#L3-L6
  - AppVeyor tests and builds "v0.71.0", then upload to (pre-existing) "v0.71.0" GitHub release
    - https://ci.appveyor.com/project/mackerelio/mackerel-agent/builds/36819108

yseto: Thank you for the detailed explanation. 🙏 
I found an implementation that used a utility (gobump) to determine the settings later in the process.
I rewrote it using it. refer d89b058

.github/workflows/go.yml
```diff
@@ -59,8 +105,148 @@ jobs:
...
+      env:
+        PLATFORM_ID: ${{ matrix.PLATFORM_ID }}
+        MSI_SUFFIX: ${{ matrix.MSI_SUFFIX }}
+
+    - name: Signing Installer
+      run: |
+        mkdir $env:RUNNER_TEMP/sign/
+        $bytes = [Convert]::FromBase64String("${{ secrets.WINDOWS_CERT_P12_BASE64 }}")
+        [IO.File]::WriteAllBytes("$env:RUNNER_TEMP/sign/cert.p12", $bytes)
+        $SIGNTOOL = "C:/Program Files (x86)/Windows Kits/10/bin/10.0.19041.0/x86/signtool.exe";
```
astj, line 236: Hmm, it looks like this path will change when Windows SDK will be updated... https://developer.microsoft.com/ja-jp/windows/downloads/sdk-archive/
Is there any stable path for this tool?

yseto: Every time the Windows SDK is added, it seems to be installed in the environment, and it seems to be fixed in this version. 

At the moment, there are 6 versions of the WIndows10 SDK installed in windows-2019 environment.
https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md#workloads-components-and-extensions

.github/workflows/386-environment/action.yml
```diff
@@ -0,0 +1,31 @@
...
+  - name: Set PATH for MINGW32
+    shell: pwsh
+    run: |
+      echo "$env:RUNNER_TEMP/msys/msys64/mingw32/bin/" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
+
+  - name: Setup golang x86
+    shell: pwsh
+    run: |
+      $ProgressPreference = 'SilentlyContinue'
+      Invoke-WebRequest -Uri https://golang.org/dl/go1.14.13.windows-386.zip -OutFile ~/go.zip
```
astj, line 20: Maybe it's useful if this action takes go version as parameter?

astj (approved):

.github/workflows/386-environment/action.yml
yseto: Indeed, I added parameter support at https://github.com/mackerelio/mackerel-agent/pull/684/commits/ed5877a7032cf9e58e38c3bb27b678b5f6ab1ca5

PR #699, opened 2021-03-05
fix option at repository-dispatch

yseto:
I made a mistake in specifying the option


lufia (approved):
I see. repository-dispatch action will be set the target to current repository by default.

PR #710, opened 2021-05-31
[ci] replace gobump

yseto:
This PR is refines a bit the workflows.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -91,7 +91,7 @@ jobs:
     name: Build (Windows)
     runs-on: windows-2019
     needs: test-windows
-    if: github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/v')
+#   if: github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/v')
     strategy:
       matrix:
         GOARCH: ['amd64', '386']
@@ -164,7 +164,7 @@ jobs:
     name: Make Installer (Windows)
     runs-on: windows-2019
     needs: build-windows
-    if: github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/v')
+#   if: github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/v')
     strategy:
       matrix:
         include:
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -180,18 +180,11 @@ jobs:
         name: windows-build-artifacts-${{ matrix.PLATFORM_ID }}
         path: build/
 
-    - run: go get github.com/x-motemen/gobump/cmd/gobump
+    - run: node _tools/parse_version.js
+      id: parse_version
 
     - name: Build Installer
-      run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/go.yml
```diff
@@ -164,7 +164,7 @@ jobs:
     name: Make Installer (Windows)
     runs-on: windows-2019
     needs: build-windows
-    if: github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/v')
+#   if: github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/v')
```
yseto, line 167: TODO remove this comments.

_tools/parse_version.js
```diff
@@ -0,0 +1,17 @@
+const fs = require("fs/promises");
+
+async function parse() {
```
shibayu36, line 3: [may] you can use following code to track errors and to write simply

```javascript
(async function main() {
   ...
})().catch(error => {
  console.error(error);
  process.exit(1);
});

File: _tools/parse_version.js [hunk 3]
```diff
@@ -0,0 +1,17 @@
...
+  const content = await fs.readFile("version.go", "utf8");
+  const regex = new RegExp(`^const *version *= *"(?<version>.*)?"`, "m");
+  const version = regex.exec(content)?.groups?.version;
+  if (version === undefined) {
+    process.on("exit", function() {
```
shibayu36, line 8: [question] why do you use `process.on` instead of just writing process.exist directly?

_tools/parse_version.js
```diff
@@ -0,0 +1,17 @@
+const fs = require("fs/promises");
+
+async function parse() {
+  const content = await fs.readFile("version.go", "utf8");
+  const regex = new RegExp(`^const *version *= *"(?<version>.*)?"`, "m");
```
shibayu36, line 5: `(?<version>.+)` is better?

shibayu36:
```suggestion
  const regex = new RegExp(`^const +version *= *"(?<version>.+)?"`, "m");
```

.github/workflows/go.yml
```diff
@@ -180,18 +180,11 @@ jobs:
         name: windows-build-artifacts-${{ matrix.PLATFORM_ID }}
         path: build/
 
-    - run: go get github.com/x-motemen/gobump/cmd/gobump
+    - run: node _tools/parse_version.js
```
shibayu36, line 183: Is there any guarantee that `node` command is installed?  I think this workflow file is for `go`.

shibayu36 (approved):
LGTM!

.github/workflows/go.yml
yseto: node is doing installed on [virtual environment](https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md).

.github/workflows/go.yml
yseto: I did it on https://github.com/mackerelio/mackerel-agent/pull/710/commits/bb9f84c0b605873d8f007fbd4748281a91fc0359

yseto:
thanks for the review!

PR #740, opened 2021-11-09
spec: reuse http.Client

lufia:
The document describes:

> Transports should be reused instead of created as needed. Transports are safe for concurrent use by multiple goroutines.


yseto (approved):
seems good.

PR #786, opened 2022-06-21
add s, ms, bps to metric units

Arthur1:
The following metric units are now available:

- seconds [s]
- milliseconds [ms]
- bits/sec [bps]

Then we added to the comment about metric units.


yseto (approved):
seems good.

PR #792, opened 2022-07-10
Loosen the conditions of delaying report of check monitering.

sugy:
- Change the log level when delaying report of check monitering.
  - I want to know if the report is delayed, even if it's not DEBUG level.
- Loosen the conditions of delaying report of check monitering.
  - In my environment, the timing when the number of reports is greater than the number of check monitors occurs once or twice an hour.
  - If the number of check monitoring is high and the number of reports is high, waiting in 30 seconds is fatal, and the number of reports will not decrease. The number of reports will continue to increase.

### DIFF
--- a/command/command.go
+++ b/command/command.go
@@ -536,7 +536,7 @@ func runCheckersLoop(ctx context.Context, app *App, termCheckerCh <-chan struct{
 		// Extend the delay when there are lots of reports
 		if len(reports) > len(app.Agent.Checkers) {
 			reportCheckDelay = reportCheckDelaySecondsMax
-			logger.Debugf("RunCheckerLoop: Extend the delay to %d seconds. There are %d reports.", reportCheckDelay, len(reports))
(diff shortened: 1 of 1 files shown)


sugy:
Hello, @Arthur1. thanks comments.

Your suggested is simpler!
I will change it that way because I can still achieve my objectives with what you have suggested.

sugy:
@Arthur1 I tried to fix it. How is this?

Arthur1:
@sugy 

Thank you! Please wait a bit longer for the review to be completed.

Arthur1 (approved):
It looks great to us. I made sure the program worked.

Arthur1:
We now recognize that the procedure of step-wisely sending check monitoring reports is so complex.

Fixing this problem requires an entire refactorization, and it takes longer. While we work out a suitable mechanism, the check reports accumulate. 

We accept your changes to resolve the sending problem quickly. Thanks for your contribution, @sugy !

sugy:
Thank you for accepting my request. 😄

Arthur1:
Hello, @sugy.

The new version of mackerel-agent, which includes your changes, has been released!

https://github.com/mackerelio/mackerel-agent/releases/tag/v0.73.0

PR #793, opened 2022-07-16
Add `omitted_success_message` option for check plugins

sugy:
- If `omitted_success_message` option is true, overwrite the success message with a fixed message and omit it.
- This option make mackerel-agent not report the OK status of the plugin to the server unless it is changed from OK status.


sugy:
I solved this implementation by wrapping the check command in a shell script. :bow:

PR #805, opened 2022-09-01
replace io/ioutil

yseto:
io/ioutil is obsolete and will be replaced

### DIFF
--- a/command/command_test.go
+++ b/command/command_test.go
@@ -3,7 +3,6 @@ package command
 import (
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
 	"os"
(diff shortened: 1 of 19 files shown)


command/command_test.go
```diff
@@ -90,7 +89,7 @@ func newMockAPIServer(t *testing.T) (config.Config, map[string]func(*http.Reques
 		fmt.Fprint(w, string(respJSON))
 	}))
 
-	root, err := ioutil.TempDir("", "mackerel-agent-test")
+	root, err := os.MkdirTemp("", "mackerel-agent-test")
```
lufia, line 92: I think it is more simple to use `testing.T.TempDir` that both `ioutil.TempDir` and following `os.RemoveAll`.

lufia (approved):
Thanks~

PR #807, opened 2022-09-01
In the test, if the Fatal if the result is nil.

yseto:
In the test, if the Fatal if the result is nil.

- possible nil pointer dereference (SA5011)
- this check suggests that the pointer can be nil


ne-sachirou (approved):
✅

PR #811, opened 2022-09-05
config_test: Add the case of LoadConfigWithInvalidToml

wafuwafu13:
Testing for errors that occur here.
https://github.com/mackerelio/mackerel-agent/blob/6ebc41b626a83361c0b5bdd6f1c1e5eb300aa161/config/config.go#L485

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -601,7 +601,7 @@ var tomlQuotedReplacer = strings.NewReplacer(
 )
 
 func TestLoadConfigFileInclude(t *testing.T) {
-  configDir := t.TempDir()
+	configDir := t.TempDir()
 	t.Cleanup(func() { os.RemoveAll(configDir) })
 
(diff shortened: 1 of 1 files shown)


config/config_test.go
```diff
@@ -186,6 +186,26 @@ func TestLoadConfigWithInvalidIgnoreRegexp(t *testing.T) {
...
+[plugin.metrics.mysql
+command = "ruby /path/to/your/plugin/mysql.rb"
+`
+
+func TestLoadConfigWithInvalidToml(t *testing.T) {
+	tmpFile, err := newTempFileWithContent(sampleConfigWithInvalidToml)
+	if err != nil {
+		t.Errorf("should not raise error: %v", err)
+	}
+	defer os.Remove(tmpFile.Name())
```
lufia, line 201: Could we use `testing.T.Cleanup` instead?

```go
t.Cleanup(func(){
    os.Remove(tmpFile.Name())
})
```

lufia (approved):

Issue #810, opened 2022-09-05
Improve `mackerel-agent configtest`


wafuwafu13:
Now, we [run `resolveConfig`](https://github.com/mackerelio/mackerel-agent/blob/6ebc41b626a83361c0b5bdd6f1c1e5eb300aa161/commands.go#L102-L103) to do configtest and detect Invalid syntax ([config_test.go](https://github.com/mackerelio/mackerel-agent/blob/master/config/config_test.go) shows that).
But, we can't detect unexpected key, so I want to improve by using [Undecoded()](https://pkg.go.dev/github.com/BurntSushi/toml#MetaData.Undecoded).
It would be better if it could suggest intended key.

This is sample of `mackerel-agent.conf` which includes unexpected key.
```
1  apikey = "abcdefg"
2  podfile = "/path/to/pidfile"
3 
4  [foo]
5  command = "bar"
6 
7  [plugins.checks.foo]
8  command = "bar"
9 
10 [plugin.check.foo]
11 command = "bar"
```

## Actual behavior

```zsh
$ mackerel-agent configtest
/path/to/mackerel-agent.conf Syntax OK
```

## Expected behavior

```
$ mackerel-agent configtest
warn: podfile is invalid key. Do you mean pidfile ?(line 2)
warn: foo is invalid key (line 4)
warn: plugins.checks.foo is invalid key. Do you mean plugin.checks.foo ? (line 7)
warn: plugin.check.foo is invalid key. Do you mean plugin.checks.foo ? (line 10)
```

### Implementation image

Create `config/validate.go#ValidateConfigFile` and use it in `commands.go#doConfigtest`

```go
func doConfigtest(fs *flag.FlagSet, argv []string) error {
	conf, err := resolveConfig(fs, argv)
	if err != nil {
		return fmt.Errorf("failed to test config: %s", err)
	}
	warnings := config.ValidateConfigFile(conf.Conffile)
	if warnings != nil {
                 return <display warnings>
        }
        fmt.Fprintf(os.Stderr, "%s Syntax OK\n", conf.Conffile)
	return nil
}
```

### Implementation stage

- [x] detect unexpected key https://github.com/mackerelio/mackerel-agent/pull/813
- [x] show suggestion https://github.com/mackerelio/mackerel-agent/pull/818

PR #818, opened 2022-09-28
Improve `mackerel-agent configtest`: Add suggestion to unexpected keys

wafuwafu13:
part of: https://github.com/mackerelio/mackerel-agent/issues/810

my conf
```my.conf
$ cat /usr/local/etc/mackerel-agent-sample.conf
apikey = "123456abcde"
podfile = "/path/to/pidfile"

[foobar]
command = "test command"
env = { FOO = "BAR" }

[filesystems]
ignore = "/path/to/ignore"
use_mntpoint = true

[plugin.foo.bar]
command = "test command"
env = { FOO = "BAR" }

[plugin.metric.1]
command = "test command"

[plugin.check.1]
command = "test command"

[plugin.check.2]
command = "test command"

[plugins.check.1]
command = "test command"

[plugin.metrics.correct]
command = "test command"

[plugin.checks.correct]
command = "test command"

[plugin.metrics.1]
command = "test command"
action = { command = "test command", use = "test user", env = { TEST_KEY = "VALUE_1" } }

[plugin.metrics.2]
command = "test command"
action = { command = "test command", xxx = "yyy" }

[plugin.checks.1]
command = "test command"
action = { command = "test command", user = "test user", en = { TEST_KEY = "VALUE_1" } }
```

result
スクリーンショット 2022-09-28 16 00 00

### DIFF
--- a/config/validate.go
+++ b/config/validate.go
@@ -24,10 +24,9 @@ func normalizeKeyname(f reflect.StructField) string {
 	return name
 }
 
-func addKeynameToCandidates(f reflect.StructField, candidates *[]string) *[]string {
+func addKeynameToCandidates(f reflect.StructField, candidates []string) []string {
 	name := normalizeKeyname(f)
-	*candidates = append(*candidates, name)
-	return candidates
+	return append(candidates, name)
 }
 
 func makeCandidates(t reflect.Type) []string {
@@ -41,12 +40,12 @@ func makeCandidates(t reflect.Type) []string {
 			continue
 		}
 		if s := f.Tag.Get("conf"); s == "parent" {
-			addKeynameToCandidates(f, &candidates)
+			candidates = addKeynameToCandidates(f, candidates)
 			childCandidates := makeCandidates(f.Type)
 			candidates = append(candidates, childCandidates...)
 			continue
 		}
-		addKeynameToCandidates(f, &candidates)
+		candidates = addKeynameToCandidates(f, candidates)
 	}
 
 	return candidates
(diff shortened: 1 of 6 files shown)


config/validate.go
```diff
@@ -3,24 +3,116 @@ package config
...
-func ValidateConfigFile(file string) ([]string, error) {
+func ValidateConfigFile(file string) ([]UnexpectedKey, error) {
 	config := &Config{}
 	md, err := toml.DecodeFile(file, config)
 	if err != nil {
 		return nil, fmt.Errorf("failed to test config: %s", err)
 	}
 
-	var unexpectedKeys []string
+	var suggestions = []string{
```
lufia, line 35: *reflect* package could get exported keys of *Config* type.
https://go.dev/play/p/BO67RHUMeXg

config/validate.go
```diff
@@ -3,24 +3,123 @@ package config
...
+	if s := f.Tag.Get("toml"); s != "" {
+		name = s
+	}
+	return name
+}
+
+func addKeynameToCandidates(f reflect.StructField, candidates *[]string) *[]string {
+	name := normalizeKeyname(f)
+	*candidates = append(*candidates, name)
+	return candidates
```
lufia, line 30: Are there a reason why this function receives a pointer to slice? In my opinion, a pointer to a slice is rare appearing in the Go code. So, maybe it is sufficient to return a slice:

```go
func addKeynameToCandidates(..., candidate []string) []string {
    ...
    return append(candidates, name)
}
```

and the receiver side:

```go
candidates = addKeynameToCandidates(f, candidates)
```

lufia (approved):
Thanks!

Issue #829, opened 2022-11-05
Contents of suggestion when executing configtest may not be correct


ryuichi1208:
If you create a config file like the one below and run configtest, a suggestion will be output to warnings. However, I could not understand the meaning of this content. Is this correct behavior?

```
apikey = "aaa"

[plugin.metrics.aaa]
command = ["exit 1"]
[plugin.metrics.aaa.bbb] ← Target of warning
command = ["exit 2"]
```

```
$ mackerel-agent version
mackerel-agent version 0.73.2 (rev 3fd96af) [darwin amd64 go1.19]

$ mackerel-agent configtest -conf a.conf
[WARNING] plugin.metrics.aaa.bbb is unexpected key.
[WARNING] plugin.metrics.aaa.bbb.command is unexpected key. did you mean plugin.metrics.aaa.bbb.command ?
```

Arthur1:
@ryuichi1208 

As you pointed out, this suggestion in the error message is inappropriate. Thank you for letting us know! Your Pull Request will also be reviewed later.

Arthur1:
@ryuichi1208 [mackerel-agent v0.73.3](https://github.com/mackerelio/mackerel-agent/releases/tag/v0.73.3)  including your pull request has been released. Thank you for your contribution!

Arthur1:
@wafuwafu13 Can you work on the refactoring for configtest as a separate issue? We close this issue.

PR #830, opened 2022-11-05
Fix config test

ryuichi1208:
refs: https://github.com/mackerelio/mackerel-agent/issues/829

```
# befor
[WARNING] plugin.metrics.aaa.action.max_check_attempts is unexpected key. did you mean plugin.metrics.aaa.action.max_check_attempts ?

# after
[WARNING] plugin.metrics.aaa.action.max_check_attempts is unexpected key.
```

### DIFF
--- a/config/validate.go
+++ b/config/validate.go
@@ -2,12 +2,13 @@ package config
 
 import (
 	"fmt"
-	"github.com/BurntSushi/toml"
-	"github.com/agext/levenshtein"
 	"reflect"
 	"sort"
 	"strings"
 	"unicode"
+
+	"github.com/BurntSushi/toml"
+	"github.com/agext/levenshtein"
 )
 
 // UnexpectedKey represents result of validation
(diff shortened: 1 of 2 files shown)


config/validate.go
```diff
@@ -55,7 +56,9 @@ func nameSuggestion(given string, candidates []string) string {
 	for _, candidate := range candidates {
 		dist := levenshtein.Distance(given, candidate, nil)
 		if dist < 3 {
-			return candidate
+			if candidate != given {
```
Arthur1, line 59: [nit] The comparison `candidate != given` could be described succinctly using the Levenshtein distance.

Arthur1 (approved):
@ryuichi1208 

Thank you for your pull request! We merge this and release it, as this simple diff would resolve the issue #829 .

By the way, we are aware of the problem that the entire logic of key suggestion is complex. Please note that we may change the code around this in the near future.

PR #832, opened 2022-11-07
fix(configtest): don't detect child key if parent key has already detected

wafuwafu13:
related: https://github.com/mackerelio/mackerel-agent/issues/829

### DIFF
--- a/config/validate.go
+++ b/config/validate.go
@@ -90,9 +90,9 @@ func ValidateConfigFile(file string) ([]UnexpectedKey, error) {
 
 	for _, v := range undecodedKeys {
 		splitedKey := strings.Split(v.String(), ".")
-		key := splitedKey[0]
+		topKey := splitedKey[0]
 		parentKey := strings.Join(splitedKey[:len(splitedKey)-1], ".")
-		if containKeyName(parentKeys, key) {
+		if containKeyName(parentKeys, topKey) {
 			// if parent is already exists in unexpectedKeyNames, child isn't detected.
 			if containKeyName(unexpectedKeyNames, parentKey) {
 				continue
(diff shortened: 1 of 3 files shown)


config/validate.go
```diff
@@ -75,99 +73,145 @@ func ValidateConfigFile(file string) ([]UnexpectedKey, error) {
...
+			topKey -> plugins
+			lastKey -> incorrect
+			parentKey -> plugins.check
+
+			v: plugins.check.incorrect.command
+			splitedKey -> [plugins, check, incorrect, command]
+			topKey -> plugins
+			lastKey -> command
+			parentKey -> plugins.check.incorrect
+		*/
```
Arthur1, line 133: I think only a simple example of each would suffice to explain. Were the procedure so complex that all test cases must be exemplified, it would have to be divided into functions with unit tests.

config/validate.go
```diff
@@ -75,99 +73,145 @@ func ValidateConfigFile(file string) ([]UnexpectedKey, error) {
...
-			if suggestName == "" {
-				unexpectedKeys = append(unexpectedKeys, UnexpectedKey{
-					v.String(),
-					"",
-				})
+		topKey := splitedKey[0]
+		lastKey := splitedKey[len(splitedKey)-1]
+		parentKey := strings.Join(splitedKey[:len(splitedKey)-1], ".")
+		// when parentKey ex) plugin.checks.incorrect.action.en or topKey ex) plugins is already exists in detectedKeys,
+		// childKey ex) plugin.checks.incorrect.action.en.TEST_KEY, plugins.check.incorrect.command isn't detected.
```
Arthur1, line 139: I recommend another expression:

> When parentKey (e.g., plugin.checks.incorrect.action.en) or topKey (e.g., plugins) already exists in detected keys,

Arthur1 (commented):
The new procedures seem to be great. Please see the suggestions about code comments.

Arthur1 (approved):
Looks greatto us.

Issue #836, opened 2022-11-28
better to remove id file at purge stage (Debian package)


kmuto:
Hi,
I discarded the old organization and purged the mackerel-agent package 0.73.3-1.systemd from my Debian environment. Later, when I installed for the new organization, I got an error that the id file is different.

I noticed mackerel-agent Debian package wouldn't remove `/var/lib/mackerel-agent/id` file even if I ran `apt-get purge mackerel-agent`.

Of course `remove` should keep configuration and associtated files, but as far as I know, `purge` is required to remove files automatically generated by the package, as well as the package's own files .

I think it is needed to add `rm -f /var/lib/mackerel/id` in`purge` condition of postrm.

https://github.com/mackerelio/mackerel-agent/blob/5197dedfe9e9f48a734d7741b9c5ca259ea55aa5/packaging/deb/debian/mackerel-agent.postrm#L4-L7
 
I'm not sure about RPM package.

azukiazusa1:
Thank you for feedback!
As you say, mackerel-agent should remove `/var/lib/mackerel-agent/id` file if you ran `apt-get purge mackerel-agent`.

Would it be possible to submit a pull request?

PR #837, opened 2022-12-04
improve purge stage of Debian package to remove id file and keep custom files

kmuto:
For #836

Here are two fixes for Debian/Ubuntu package.

- remove `/var/lib/mackerel-agent/id` file at purge stage.
- remove `/etc/mackerel-agent/mackerel-agent.conf` explicitly instead of `rm -rf /etc/mackerel-agent`, and define `var/lib/mackerel-agent` as package own directory.

Second change is intended to keep user's file in `/etc/mackerel-agent` and `/var/lib/mackerel-agent`. The purge stage should remove the files what package created,  but should not remove user's own files.

[normal]
```
$ ls /etc/mackerel-agent /var/lib/mackerel-agent
/etc/mackerel-agent:
mackerel-agent.conf  mackerel-agent.conf.example

/var/lib/mackerel-agent:
id
$ sudo apt-get purge mackerel-agent
 ...
Removing mackerel-agent (0.73.3-1.systemd) ...
(Reading database ... 688477 files and directories currently installed.)
Purging configuration files for mackerel-agent (0.73.3-1.systemd) ...
$ ls /etc/mackerel-agent /var/lib/mackerel-agent
ls: cannot access '/etc/mackerel-agent': No such file or directory
ls: cannot access '/var/lib/mackerel-agent': No such file or directory
```

[with user's custom file]
```
$ ls /etc/mackerel-agent /var/lib/mackerel-agent
/etc/mackerel-agent:
mackerel-agent.conf  mackerel-agent.conf.example  my-note.txt

/var/lib/mackerel-agent:
id my-note.txt

$ sudo apt-get purge mackerel-agent
 ...
Purging configuration files for mackerel-agent (0.73.3-1.systemd) ...
dpkg: warning: while removing mackerel-agent, directory '/var/lib/mackerel-agent' not empty so not removed
dpkg: warning: while removing mackerel-agent, directory '/etc/mackerel-agent' not empty so not removed
$ ls /etc/mackerel-agent /var/lib/mackerel-agent
/etc/mackerel-agent:
my-note.txt

/var/lib/mackerel-agent:
my-note.txt
```

### DIFF
--- a/packaging/deb-systemd/debian/mackerel-agent.dirs
+++ b/packaging/deb-systemd/debian/mackerel-agent.dirs
@@ -1,2 +1,3 @@
 usr/bin
 etc/mackerel-agent
+var/lib/mackerel-agent
--- a/packaging/deb-systemd/debian/mackerel-agent.postrm
+++ b/packaging/deb-systemd/debian/mackerel-agent.postrm
@@ -2,7 +2,8 @@
 set -e
 
 if [ "$1" = "purge" ]; then
-  rm -rf /etc/mackerel-agent
+  rm -f /etc/mackerel-agent/mackerel-agent.conf
+  rm -f /var/lib/mackerel-agent/id
 fi
 rm -f /var/run/mackerel-agent.pid
 


ne-sachirou (approved):
❤️

PR #869, opened 2023-02-09
fix msi package installscope.

yseto:
We need `InstallScope="perMachine"`


Arthur1 (approved):
It looks great to us.

PR #898, opened 2023-05-09
add feature to ignore disk labels by regexp in 'disk' of system metrics

kmuto:
This patch aims to allow the disk label to be ignored using Go regular expressions, as well as filesystems and interfaces.

example of mackerel-agent.conf:
```
[disks]
ignore = "loop.*"
```

(Unfortunately, the system metric graphs cannot be rebuilt, so users have to write above settings before running mackerel-agent to avoid extra disk labels appearing in the graph.)

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -94,7 +94,7 @@ jobs:
     name: Build (Windows)
     runs-on: windows-2019
     needs: test-windows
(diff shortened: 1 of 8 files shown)


kmuto:
I checked it worked on Debian GNU/Linux 64bit and Windows 11 Professional 64bit as expected.

inommm (approved):
LGTM

PR #927, opened 2023-08-10
remove comment that differ from current plugin implementation

kga:
Currently, mackerel-plugin-redis supports password authentication. refs. https://github.com/mackerelio/mackerel-agent-plugins/pull/232


yseto (approved):
seems good.

PR #947, opened 2023-11-16
Avoid iowait percentage's overflow when counter is reset

Arthur1:
IOwait counter retrieved from /proc/stat can be reset.

> I noticed that the "iowait" mode was also the only mode which had any counter resets.
> https://github.com/prometheus/prometheus/issues/2856#issue-236546854

I created a function to calculate the difference of counters considering resets to avoid overflows.


ne-sachirou (approved):
🙆‍♀️

PR #963, opened 2024-01-17
introduce monitorDelaySeconds to wait for user's specific services that take a long time to start up

kmuto:
The mackerel-agent may raise an alert for monitored services that take a long time to start up.
Here is proposed to introduce an option called `monitorDelaySeconds` for the wait seconds to start monitoring.

- An upper limit of around 3600 seconds would be appropriate, but since other settings of the agent did not seem to have a detailed limit, I just have set it to the uint16 range.
- The waiting time addition using host IDs is left as it is.
- The go routine for updating host meta information is executed without waiting.
- There will also be a wait time when restarting, but this is acceptable, I believe.

### DIFF
--- a/command/command.go
+++ b/command/command.go
@@ -212,7 +212,7 @@ func loop(app *App, termCh chan struct{}) error {
 	go enqueueLoop(ctx, app, postQueue)
 
 	postDelaySeconds := delayByHost(app.Host)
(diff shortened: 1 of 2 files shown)


kmuto:
I withdraw this PR because this is too ad-hoc, not solution.

PR #983, opened 2024-03-07
use Go 1.22 only Unix-like OSes builds

lufia:
* upgrade Go version to 1.22.x
* split out build steps as a reusable workflow from **go.yml**


Arthur1:
見ますが時間かけさせてください。やりすぎな印象があって色々言語化していくのに時間がかかる

.github/workflows/go-build.yml
```diff
@@ -0,0 +1,50 @@
+name: Build(Go)
+run-name: Build (${{ inputs.os-version }}/${{ inputs.go-version }})
+
+on:
+  workflow_call:
+    inputs:
+      os-version:
+        description: 'GitHub-hosted runnners'
+        default: ubuntu-22.04
```
ne-sachirou, line 9: I want the default value to be well without maintenance than parameters passed from outside, I want it to be stable. I think `ubuntu-latest` is more stable than `ubuntu-22.04`.

PR #994, opened 2024-04-11
Ignore errors of getting information of a disk on Windows

Arthur1:
In the Windows version, if access to ONE disk is denied, information and metrics cannot be retrieved FOR ALL disks.

I have changed it so that if an access error occurs for a disk, retrieving information for only that disk is skipped.


lufia (approved):
:ok_person:

PR #1000, opened 2024-05-01
list wix packages

ne-sachirou:
https://github.com/mackerelio/mackerel-agent/actions/runs/8904749775/job/24454574171
>Error: plugins_windows.go:7:2: import "github.com/mackerelio/go-check-plugins/check-disk" is a program, not an importable package

I checked that the `build-windows` jobs are succeeded.
- https://github.com/mackerelio/mackerel-agent/actions/runs/8906052482/job/24457719885?pr=1000
- https://github.com/mackerelio/mackerel-agent/actions/runs/8906052482/job/24457720029?pr=1000

.

- cf. https://github.com/golang/go/issues/59186

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -62,7 +62,7 @@ jobs:
     name: Build (Windows)
     runs-on: windows-2022
     needs: test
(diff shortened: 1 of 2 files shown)


yseto (approved):
seems good.

PR #1021, opened 2024-07-08
Use Azure Trusted Signing for Windows installer code signing

yohfee:
Use Azure Trusted Signing to keep Windows installer code signing through CI.


yseto (approved):
seems good.

PR #1025, opened 2024-11-18
fix lint error

rmatsuoka:
Linter have said "printf: non-constant format string in call to fmt.Errorf (govet)"

I just replaced fmt.Errorf into errors.New

https://github.com/mackerelio/mackerel-agent/actions/runs/11892185199/job/33134502506

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -3,6 +3,7 @@
 package main
 
 import (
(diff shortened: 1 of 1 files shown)


yohfee (approved):
👍

PR #1027, opened 2024-11-22
add mackerel-plugin-snmp and check-ping to Windows build.

kmuto:
There is a particular demand for monitoring networks centred around Windows Server, especially on-premises.
The aim of this PR is to bundle mackerel-plugin-snmp and check-ping, which are often used with Linux for an network administration, into the Windows version of the mackerel-agent installer.

I have verified that mackerel-plugin-snmp and check-ping work correctly on Windows using Windows Server 2022.


yseto (approved):
seems good.

PR #1087, opened 2025-05-19
Updated runner to use latest Windows 2025 image

appare45:
I'll check if the installer built on Windows Server 2025 works after merging into main.


yseto (approved):
seems good.

appare45:
I've checked that both mackerel-agent.msi and mackerel-agent-x64.msi work.

PR #1107, opened 2025-09-08
use Go 1.24.x

yseto:
I did.


yseto:
Thanks for your review.

PR #1223, opened 2026-06-17
Use the time reported by the plugin.

yseto:
We will use the time reported by the plugin.
mackerel-agent uses current of the time of aggregation. Time information for delayed (or future) metrics was not used.
This fix improves accuracy.


command/command.go
```diff
@@ -407,20 +407,25 @@ func enqueueLoop(ctx context.Context, app *App, postQueue chan *postValue) {
 						continue
 					}
 				}
-				for name, value := range values.Values {
-					if math.IsNaN(value) || math.IsInf(value, 0) {
-						logger.Warningf("Invalid value: hostID = %s, name = %s, value = %f\n is not sent.", hostID, name, value)
+				for name, attribute := range values.Values {
+					if math.IsNaN(attribute.Value) || math.IsInf(attribute.Value, 0) {
+						logger.Warningf("Invalid value: hostID = %s, name = %s, value = %f\n is not sent.", hostID, name, attribute)
```
ne-sachirou: When passing attribute.Value to the invalid value log.
When detecting NaN/Inf metric values, this log currently passes the entire `metrics.ValueAttribute` structure to `%f`, resulting in format errors like `%!f(...)` instead of the actual value. This prevents reading the required values for investigating invalid values, so `attribute.Value` should be passed here instead.

NaNやInf が渡された場合、このログは `%f` に `metrics.ValueAttribute` 構造体を渡すため、実際の値ではなく `%!f(...)` のようなフォーマットエラーを出力します。無効な値の調査に必要な値が読めなくなるので、ここは `attribute.Value` を渡してください。

yseto: I fixed at b3b5f798bd9a376823a3b687257ec67675231a9f

ne-sachirou (approved):
🙆‍♀️

yseto:
Thanks for your review.

PR #1254, opened 2026-09-16
KillMode=control-group (default) was conflict the processing on_stop

yseto:
When KillMode=mixed is set, the signal is sent only to the parent process when the service is stopped.


ne-sachirou (approved):
🙆‍♀️

yseto:
Thanks for your review.

